PR #2, opened 2026-01-30
(docker) fix platform on manylinux wheel

zouzias:
Explicitly passes the `--plat manylinux_2_34_x86_64` parameter in `auditwheel repair` to make sure the output wheel has the correct tag `ptoas-0.1.0-py3-none-manylinux_2_34_x86_64.whl`.

Generated wheel is already released on pypi: https://pypi.org/project/ptoas/#files

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -41,7 +41,7 @@ RUN ninja -C $LLVM_BUILD_DIR
 
 # build ptoas
 WORKDIR $WORKSPACE_DIR
-RUN git clone https://github.com/learning-chip/PTOAS.git
+RUN git clone https://github.com/zhangstevenunity/PTOAS.git
 # TODO: tag git commit of PTOAS repo
 
 WORKDIR $PTO_SOURCE_DIR


docker/Dockerfile
```diff
@@ -74,7 +74,7 @@ RUN python setup.py bdist_wheel
 
 # fix missing so
 RUN export LD_LIBRARY_PATH=$LLVM_BUILD_DIR/lib:$PTO_INSTALL_DIR/lib:$LD_LIBRARY_PATH \
-    && auditwheel repair dist/ptoas*.whl
+    && auditwheel repair --plat manylinux_2_34_x86_64 dist/ptoas*.whl
```
learning-chip: This dockerfile also works for aarch64, need a top level `ARG` for the arch suffix

zouzias: Fixed in https://github.com/zhangstevenunity/PTOAS/pull/2/changes/df2cb5a9298a5eb9900248eaa4ea5b643a2a9684

PR #9, opened 2026-01-31
Verify across-platform compatibility of python wheel and ptoas binary, using multi-stage docker build 

learning-chip:
- The final docker image only contains: `ptoas-*.whl`, `ptoas` binary, and the minimum `*.so` dependencies needed (less than 200 MB total). Ensure unit tests still work without access to full llvm build dir. 
- Tested e2e pipeline during docker build: python binding -> `ptoas` -> `bisheng`
- Use different linux variants (CentOS for build, ubuntu for runtime) to ensure the wheel is portable.
- Tested on x86 and ARM host, both work (just change `ARG ARCH`)

Ready for review & merge @zhangstevenunity @zouzias 

Can be further improved in future PRs:
1. let `ptoas` static linked to avoid copying a directory of `*.so` dependencies 
2. add `ptoas` as part of wheel, so that `pip install` and also bring this binary cli
(see my code comments below for why those improvements didn't work yet in current PR)

TODO for user's convenience:
- Upload latest wheel to https://pypi.org/project/ptoas/ (@zouzias )
- Upload latest `ptoas` to https://github.com/zhangstevenunity/PTOAS/releases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,65 @@
+# Build artifacts
+build/
+install/
+cmake-build-*/
+CMakeFiles/
+CMakeCache.txt
+CMakeScripts/
+Testing/
+Makefile
+cmake_install.cmake
+install_manifest.txt
+CTestTestfile.cmake
+compile_commands.json
+*.ninja
+*.ninja_log
+*.ninja_deps
+
+# Binary outputs
+*.o
+*.obj
+*.lo
+*.la
+*.a
+*.lib
 *.so
+*.dylib
+*.dll
+*.exe
+*.out
+*.app
+
+# Debug/profiling
+*.pdb
+*.dSYM/
+*.gcno
+*.gcda
+*.profraw
+*.profdata
+
+# Python
+__pycache__/
+*.py[cod]
+*.pyd
+*.egg-info/
+.eggs/
+.venv/
+venv/
+.env
+dist/
+
+# Logs/temp
+*.log
+*.tmp
+*.swp
+*.swo
+.cache/
+.pytest_cache/
+
+# IDE/editor
+.vscode/
+.idea/
+*.iml
+.DS_Store
+.ipynb_checkpoints/
+*.orig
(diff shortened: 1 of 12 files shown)


docker/README.md
```diff
@@ -1,14 +1,50 @@
...
+sudo docker run --rm -it \
+    -v $HOME:/mounted_home -w /mounted_home \
+    ptoas:py3.11 /bin/bash
+
+cd /sources/test/Abs
+python ./abs.py > ./abs.pto
+ptoas --enable-insert-sync ./abs.pto -o ./abs.cpp
+
+# temporary fix for macro guards, better let `ptoas` insert it
+(echo '#if __CCE_AICORE__ == 220 && defined(__DAV_C220_VEC__)'; cat abs.cpp; echo '#endif') > abs.cpp.tmp && mv abs.cpp.tmp abs.cpp
```
learning-chip: The output cpp file from `ptoas` does not contain this `defined(__DAV_C220_VEC__)` guard. Is there a way to avoid this ad-hoc patching? @zhangstevenunity

docker/README.md
```diff
@@ -1,14 +1,50 @@
...
+    --device=/dev/davinci4 --device=/dev/davinci5 \
+    --device=/dev/davinci6 --device=/dev/davinci7 \
+    --device=/dev/davinci_manager \
+    --device=/dev/devmm_svm \
+    --device=/dev/hisi_hdc  \
+    -v /usr/local/bin/npu-smi:/usr/local/bin/npu-smi \
+    -v /usr/local/Ascend/driver:/usr/local/Ascend/driver:ro \
+    -v /etc/ascend_install.info:/etc/ascend_install.info:ro \
+    -v $HOME:/mounted_home -w /mounted_home \
+    ptoas:py3.11 /bin/bash
```
zouzias: `ptoas:py3.11` -> `ptoas:py3.12`?

PR #55, opened 2026-02-06
Block index relu

filipzrc:


test/kernel_launch/tiled_relu/run.py
```diff
@@ -0,0 +1,56 @@
...
+    default_stream_ptr = torch.npu.current_stream()._as_parameter_
+
+    def relu_func(
+        x,
+        y,
+        block_dim=default_block_dim,
+        stream_ptr=default_stream_ptr
+        ):
+        print(f"streamptr: {stream_ptr}")
+        lib.call_kernel(
```
learning-chip: Need to change the caller function to launch multi-core kernel @filipzrc 

You can refer to how the cpp demo launches kernel:
https://gitcode.com/cann/pto-isa/blob/master/demos/torch_jit/add/add_custom.cpp

PR #317, opened 2026-03-19
fix(emitc): elide redundant bind_tile rewrap

HecreReed:
## Summary
- elide redundant `bind_tile` EmitC rewrites when the source tile already has the exact same static tile type
- keep the existing rebind path for cases that still need new runtime-valid metadata or different configs/layouts
- add a regression test covering `alloc_tile -> tprint` so the generated C++ stays compact

## Why
For simple cases like `alloc_tile` followed by `pto.tprint`, the current lowering creates a second sibling tile and rebinds through `tile.data()` + `reinterpret_cast<uint64_t>`, even though the original tile is already the exact value `TPRINT` needs. The code works, but the generated C++ is unnecessarily noisy.

## Validation
- `cmake --build build -j8`
- `build/tools/ptoas/ptoas test/basic/tprint_alloc_tile_no_rebind.pto | FileCheck test/basic/tprint_alloc_tile_no_rebind.pto`
- `build/tools/ptoas/ptoas test/basic/set_validshape_local_lowering.pto | FileCheck test/basic/set_validshape_local_lowering.pto`
- `build/tools/ptoas/ptoas test/basic/set_validshape_if.pto | FileCheck test/basic/set_validshape_if.pto`

### DIFF
--- a/lib/PTO/Transforms/PTOToEmitC.cpp
+++ b/lib/PTO/Transforms/PTOToEmitC.cpp
@@ -6802,6 +6802,12 @@ struct PTOTrapOpToEmitC : public OpConversionPattern<pto::TrapOp> {
 struct PTOBindTileToEmitC : public OpConversionPattern<pto::BindTileOp> {
   using OpConversionPattern::OpConversionPattern;
 
+  struct TileBuildSpec {
+    std::string tileTypeStr;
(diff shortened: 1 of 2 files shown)


zhangstevenunity:
@codex

PR #348, opened 2026-03-24
[WIP] Tile-native pointer_cast and PlanMemory rewiring (memref bridge removal phase)

TaoTao-real:
## Summary
This WIP PR advances the memref->tilebuf migration on the clean branch by making the address-binding path tile-native.

### What changed
- `pto.pointer_cast` result type switched from `memref` to `!pto.tile_buf`.
- Main pipeline no longer runs `PTOViewToMemref` in `tools/ptoas/ptoas.cpp`.
- PlanMemory rewrite switched from `memref.alloc -> pointer_cast` to `alloc_tile -> pointer_cast`.
- Added tile-native semantic extraction module:
  - `lib/PTO/Transforms/TileBufferSemantics.h`
  - `lib/PTO/Transforms/TileBufferSemantics.cpp`
- Updated Sync translator pointer-cast memory info to consume tilebuf result types.
- EmitC path updated:
  - `alloc_tile -> pointer_cast` pre-lowering
  - `pointer_cast` conversion consumes tilebuf result descriptors
  - view-like WIP bridging: `subset/bitcast/treshape -> pointer_cast` (currently zero-offset subset only)
- Added/updated tests and samples for tilebuf + PlanMemory path.

## Why
The current architecture goal is full-chain tilebuf semantics, avoiding memref as the central storage descriptor in PlanMemory and downstream lowering. This PR is the first WIP step that makes pointer binding and memory planning tile-native.

## Tests run
- Build:
  - `ninja -C build ptoas`
- Basic coverage:
  - `test/basic/tilebuf_semantic_smoke.pto`
  - `test/basic/tilebuf_root_trace.pto`
  - `test/basic/tilebuf_manual_addr_preserve.pto`
  - `test/basic/tilebuf_auto_addr_assign.pto`
  - `test/basic/set_validshape_local_lowering.pto`
- Sample regression subset:
  - `test/samples/runop.sh -t planmemory`

## Known limitations (WIP)
- subset->pointer_cast bridging currently supports only zero-offset subset.
- Legacy memref-specific scenarios relying on old bridge behavior are not fully ported in this PR.
- Sync/multibuffer full migration is intentionally out of scope for this phase.

## Follow-ups
- Complete non-zero-offset subset address rebasing for tile-native view lowering.
- Remove remaining memref compatibility fallbacks in PlanMemory/EmitC.
- Sync pipeline full tile-native migration.


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #348 [WIP] Tile-native pointer_cast and PlanMemory rewiring (memref bridge removal phase)
- Author: `TaoTao-real`
- Base/Head: `main` / `codex/tilebuf-replace-memref-pm-v1`
- Head SHA: `78fd5286affe`
- Trigger: 检测到新的 open PR
- Generated At: `2026-04-30T07:59:48Z`
- Status: completed

### Summary

PR #348 introduces two P1 EmitC regressions in tile view lowering, plus a P2 InsertSync alias-analysis bug and a P2 C API compatibility break.

### Findings

1. **P1** View-like lowering rejects \`pto.declare_tile\` roots `lib/PTO/Transforms/PTOToEmitC.cpp`:`303`

`tracePointerCastAddrsFromSource()` only knows how to peel `alloc_tile`, `pointer_cast`, `subview`, `bitcast`, and `treshape`. It never recognizes `pto.declare_tile`, which is what frontend pipe lowering produces for `tpop` results. As a result, any legal IR that applies `pto.subview`, `pto.bitcast`, or `pto.treshape` to a popped tile (or any user-authored `declare_tile`) will fail in EmitC lowering with the `expects ... source from pointer_cast` match failure. Before this PR, the memref bridge lowered `declare_tile` first and these view ops still compiled.

2. **P1** Bitcast/TReshape rewrites lose dynamic valid-shape metadata `lib/PTO/Transforms/PTOToEmitC.cpp`:`3739`

`PTOBitcastToPointerCast` and `PTOTReshapeToPointerCast` rebuild the result using `materializeStaticValidDims(...)` only. When the source/result tile carries dynamic valid dims, no runtime `valid_row`/`valid_col` are forwarded into the new `pto.pointer_cast`. Later, `PointerCastConversion` falls back to the full static shape when those operands are absent, so a legal dynamic-valid `pto.bitcast`/`pto.treshape` is emitted as a full-valid tile. That changes DMA/compute extents and can overread or overwrite past the runtime-valid region. The previous bridge propagated parent valid dims instead of dropping them.

3. **P2** InsertSync subview alias ranges ignore tile physical layout `lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp`:`31`

`getStaticOffsetAndSize()` now derives `pto.subview` offsets from plain contiguous `shape` strides. That ignores tile physical layout metadata (`blayout`, boxed `slayout`, `fractal`, non-compact parent stride), even though the new EmitC lowering in the same PR has to compute those physical strides explicitly. For col-major or boxed tiles, InsertSync will record the wrong byte range for a subview, so overlapping local buffers can be treated as independent and required waits may be skipped when `--enable-insert-sync` is enabled.

4. **P2** \`mlirPTOGMTypeGet\` silently changed its public type contract `lib/CAPI/Dialect/PTO.cpp`:`657`

`mlirPTOGMTypeGet()` keeps the same exported C API name and signature, but it now returns `!pto.tensor_view` instead of the GM memref type it returned before. Downstream C/Python clients that use this helper as their "GM buffer type" will start building different IR or fail existing type checks without any source change. This is a public compatibility break and needs either a new symbol or a compatibility wrapper.

PR #353, opened 2026-03-25
InferPTOLayout: accept user ND/DN in ambiguous minor-2D cases

TaoTao-real:
Summary
- Fix InferPTOLayout conflict handling for ambiguous minor-2D layout cases (ND/DN both legal).
- When user explicitly sets ND or DN and inference picks the other one in an ambiguous case, keep the user layout instead of raising layout mismatch.
- Add lit regression: test/basic/infer_layout_ambiguous_minor2d_user_dn.pto.
 
Motivation
- Users reported infer-layout failures where both ND and DN are valid, but pass rejected user-specified layout because inferred default differed.

Design
- Extend verifyOrSetLayout with ambiguity context and allow ND<->DN mismatch only when inferLayout5D reports minor-2D ambiguity.
- Apply the same logic to both make_tensor_view and memref.reinterpret_cast paths.

Testing
- ninja -C build ptoas
- build/tools/ptoas/ptoas /tmp/wrong_618.mlir -o /tmp/wrong_618_out.cpp
- build/tools/ptoas/ptoas test/basic/infer_layout_ambiguous_minor2d_user_dn.pto 2>&1 | FileCheck test/basic/infer_layout_ambiguous_minor2d_user_dn.pto

Risk / Rollback
- Risk: low, behavior change is scoped to explicit ND/DN ambiguity only.
- Rollback: revert this PR.

### DIFF
--- a/lib/PTO/Transforms/InferPTOLayout.cpp
+++ b/lib/PTO/Transforms/InferPTOLayout.cpp
@@ -73,6 +73,10 @@ struct ShapeStride5D {
   SmallVector<int64_t, 5> stride;
 };
 
+static bool isMinor2DLayout(Layout layout) {
+  return layout == Layout::ND || layout == Layout::DN;
+}
+
 static std::optional<ShapeStride5D> rightAlignTo5D(ArrayRef<int64_t> shape,
                                                    ArrayRef<int64_t> stride) {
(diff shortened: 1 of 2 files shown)


zhangstevenunity:
@codex

PR #395, opened 2026-03-30
fix(release): rewrite packaged macOS dylib install names

HecreReed:
## Summary
- add a post-collection pass that rewrites absolute non-system dylib references to packaged `@loader_path` references
- keep the portable dependency validation, but normalize the packaged Mach-O files before validating them
- log each rewritten install name so future packaging failures are easier to diagnose

## Why
Manual macOS wheel run `23724501658` still failed on `main` in `Build wheel (Python 3.11, aarch64)` during `Collect ptoas binary and dependencies`.

The new validation now works, and it caught a real packaging leak:
- `/Users/runner/work/PTOAS/PTOAS/ptoas-dist/lib/libLLVMSupport.dylib -> /opt/homebrew/opt/zstd/lib/libzstd.1.dylib`

That means dependency collection copied `libzstd.1.dylib`, but at least one packaged dylib still retained the original Homebrew absolute install name. This patch adds a final normalization pass over the packaged tree so those remaining absolute references are rewritten before validation.

## Validation
- `bash -n docker/collect_ptoas_dist_mac.sh`
- `git diff --check`
- inspected failing job log `69105423937` from run `23724501658`

### DIFF
--- a/docker/collect_ptoas_dist_mac.sh
+++ b/docker/collect_ptoas_dist_mac.sh
@@ -141,8 +141,52 @@ collect_dylibs() {
   done < <(otool -L "$bin" | awk 'NR>1 {print $1}')
 }
 
+packaged_dep_ref() {
+  local owner="$1"
+  local dep_base="$2"
+  case "$owner" in
+    "${PTOAS_DIST_DIR}/bin/"*)
+      echo "@loader_path/../lib/${dep_base}"
+      ;;
+    "${PTOAS_DEPS_DIR}/"*)
+      echo "@loader_path/${dep_base}"
+      ;;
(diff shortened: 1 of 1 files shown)


reedhecre:
## A3 板测成功
- 触发方式：`merged`
- 源码提交：`b9dd5f7e3f98`
- 结果汇总：OK `163` / FAIL `0` / SKIP `0`
- 日志：`/tmp/ptoas-board-monitor/logs/20260330_111407_merged_pr395.log`
- 结果 TSV：`/tmp/ptoas-board-monitor/logs/20260330_111407_merged_pr395.tsv`

Issue #428, opened 2026-04-02
[Bug] generated C++ Cube kernel hangs with version >= 0.18


MirkoDeVita98:
### Component
EmitC / Codegen (lib/PTO/Transforms/PTOToEmitC.cpp)

### Description
The generated C++ in versions >= 0.18 appears to preserve the same overall math,
tensor views, and tensor layout for this kernel, but it changes the pipeline event
synchronization (`set_flag` / `wait_flag`) and tail synchronization behavior.

The version generated by 0.17 works for this reproducer. In that working output,
the event sequence is initialized before entering the loops, and the kernel ends
with explicit waits that drain the outstanding event dependencies.

In the >= 0.18 output, the synchronization schedule is different. In the failing
variant, the first main loop begins with:

  wait_flag(PIPE_M, PIPE_MTE1, EVENT_ID2);
  wait_flag(PIPE_M, PIPE_MTE1, EVENT_ID3);

but there is no obvious pre-loop:

  set_flag(PIPE_M, PIPE_MTE1, EVENT_ID3);

before that first wait on `EVENT_ID3`.

I compared the MLIR from the older working version and the newer failing version,
and for this kernel they appear identical in substance. The generated C++ differs
mainly in synchronization/event scheduling, which suggests the regression is
introduced downstream during EmitC/codegen or sync insertion rather than in the
high-level IR.

Reproducer archive:
[bug_reproduce.zip](https://github.com/user-attachments/files/26436355/bug_reproduce.zip)

### Reproduction

To fully reproduce see: https://github.com/huawei-csl/pto-dsl/tree/main/examples/aot/fast_inverse/block_inversion

```mlir
module {
  func.func @tri_inv_block2x2_fp16(%arg0: !pto.ptr<f32>, %arg1: !pto.ptr<f16>, %arg2: !pto.ptr<f16>, %arg3: i32) {
    pto.section.cube {
      %c0 = arith.constant 0 : index
      %c1 = arith.constant 1 : index
      %c64 = arith.constant 64 : index
      %c32 = arith.constant 32 : index
      %0 = arith.index_cast %arg3 : i32 to index
      %1 = arith.subi %0, %c1 : index
      %2 = pto.get_block_idx
      %3 = arith.index_cast %2 : i64 to index
      %4 = pto.get_block_num
      %5 = arith.index_cast %4 : i64 to index
      %6 = arith.muli %5, %c64 : index
      %7 = arith.muli %3, %c64 : index
[...]

Issue #226, opened 2026-03-09
[Performance] `--enable-insert-sync` on a dynamic GEMM example generates 10% slower kernel than manual sync version


learning-chip:
### Summary

Record a practical use case where `ptoas --enable-insert-sync` still has  ~10% room for performance improvement, compared to a known manual-sync plan. 

### Background

I wrote a dynamic-shape matmul that is 2x faster than [original pto-isa `gemm_performance` example](https://gitcode.com/cann/pto-isa/tree/699931c6400b68bd19e3803eec61ad0268e2aeb0/kernels/manual/a2a3/gemm_performance) and 0.9~1.1x of [`aclnnMatmul` in CANN 8.5.0](https://www.hiascend.com/document/detail/zh/canncommercial/850/API/aolapi/context/ops-nn/aclnnMatmul.md). See [matmul_swizzle/simple_demo](https://github.com/huawei-csl/pto-dsl/tree/e7a68427ce867048ab2d43fd6847e320262bed51/examples/aot/matmul_swizzle/simple_demo) to reproduce.

The auto-sync version is [only ~100 lines of Python](https://github.com/huawei-csl/pto-dsl/blob/e7a68427ce867048ab2d43fd6847e320262bed51/examples/aot/matmul_swizzle/simple_demo/simple_matmul_builder.py#L227-L340), and reaching 90% of manual-sync is quite decent. I just wonder if the last 10% perf gap can be filled.

### Command line

```bash
ptoas --enable-insert-sync simple_matmul_auto_sync.pto -o simple_matmul_auto_sync.cpp
ptoas simple_matmul_manual_sync.pto -o simple_matmul_manual_sync.cpp
```

### Reproduction input

[pto_matmul.zip](https://github.com/user-attachments/files/25855588/pto_matmul.zip)

contains both inputs:
- `simple_matmul_auto_sync.pto`
- `simple_matmul_manual_sync.pto`

and outputs:
- `simple_matmul_auto_sync.cpp`
- `simple_matmul_manual_sync.cpp`

### Expected performance

Auto-sync should be ideally as fast as manual sync version. (or discover even faster pipelining?)

### Actual performance

Auto-sync is 5~15% slower than manual-sync, see the detailed PRs below (contains full code with kernel launch, and on-device performance measurement on 910B2:):

- Directly written in PTO-ISA C++: https://github.com/huawei-csl/pto-kernels/pull/26
[...]

PR #604, opened 2026-04-29
[ADR] InsertSync memory-phi/canonical-writer dedup architecture

TaoTao-real:
## Summary
This PR adds an ADR for architecture-level InsertSync improvement using Memory-Phi / Canonical-Writer modeling to deduplicate semantically equivalent sync chains in loop + if/else scenarios.

## Why
Current path-intersection logic merges sync *state* only, but cannot merge semantically equivalent dependency edges created from branch-local physical writers. This can lead to duplicated loop-carried chains (including extra seed/drain) and inflated sync count.

## What is included
- New design doc: `docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md`
- Architecture decisions, data model, invariants, algorithm phases, validation gates, and risk controls

## Scope
- Docs-only ADR PR
- No implementation changes in this PR

## Follow-up implementation
Implementation will be tracked in follow-up PRs, gated by regression suites including #428/#454 and issue-style workloads (#226/#233/FA).

### DIFF
--- a/docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md
+++ b/docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md
@@ -1,0 +1,277 @@
+# PTOAS InsertSync: Memory-Phi / Canonical-Writer 去重合并架构设计（ADR）
+
+- Status: Proposed
+- Date: 2026-04-29
+- Owners: PTOAS InsertSync maintainers
+- Scope: `InsertSync` 分析与生成层（不改前端 IR 语义，不改 CLI）
+
+---
+
+## 1. 背景与问题
+
+当前 `InsertSync` 在复杂控制流（尤其 `loop + if/else`）下，常出现“依赖语义等价但同步链重复”：
+
+1. 分支内有两个物理 writer（例如 `if: tmatmul` / `else: tmatmul.acc`）。
+2. 汇合后 consumer 只关心“该逻辑结果已可见”，但当前算法仍按两个物理 writer 各建一套同步边。
+3. 即便已有 `if/else` 路径同步状态取交集，交集机制只作用于 `alreadySync` 传播，不会合并“语义等价依赖边”。
+
+结果是：
+
+1. loop-carried `set/wait`、seed/drain 数量偏多。
+2. 某些 pipe-pair（如 `M -> MTE1`）在分支场景下出现双链或多链。
+3. 同步数量与手工版本差距明显，影响性能和 event-id 压力。
+
+---
+
+## 2. 设计目标
+
+1. 在 `InsertSync` 中建立 Memory-SSA 风格的逻辑写点建模（memory-phi / canonical-writer）。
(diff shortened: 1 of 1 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #604 [ADR] InsertSync memory-phi/canonical-writer dedup architecture
- Author: `TaoTao-real`
- Base/Head: `main` / `codex/adr-canonical-writer-memory-phi`
- Head SHA: `749a24a687ff`
- Trigger: 检测到新的 open PR
- Generated At: `2026-04-29T08:41:31Z`
- Status: completed

### Summary

ADR #604 has two correctness-level design gaps in the canonical dependency model and is missing a loop zero-trip regression gate in its validation plan.

### Findings

1. **P1** Canonical writer model does not cover WAR and ACC read/read hazards `docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md`:`100`

Lines 98-103 route every `MemoryDependentAnalyzer` candidate through `CanonicalWriterAnalysis`, and the core types only model `CanonicalWriterId` / `CanonicalDepEdge { producer, consumer, ... }`. The current pass still needs non-writer-source hazards: `IsMemInfoHasDependency` inserts syncs for `def/use` (WAR) and for the special cross-pipe ACC `use/use` case because those read/read sequences can fail on device. Those edges do not have a source writer, so the proposed schema cannot represent them. If implemented as written, this will either drop required syncs or miscanonicalize them through an unrelated writer.

2. **P1** CanonicalMemKey drops subview/range information and can merge unrelated slices `docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md`:`47`

The ADR builds one logical def chain per `(rootBuffer, groupId)` and the key types only keep `rootBuffer/groupId` plus scope metadata (lines 47, 82, 90, 124). The existing alias model is range-sensitive: `BaseMemInfo` carries `baseAddresses` and `allocateSize`, and overlap checks use those byte ranges before treating two accesses as the same memory. The repo already has real same-root disjoint-slice patterns such as `test/lit/pto/subview_tmatmul_acc_slices_a5.pto`. Unless `groupId` is explicitly defined to encode offset/size-equivalence, BranchPhi/LoopPhi can merge unrelated slices of the same root buffer and dedup away syncs that are still required.

3. **P2** Validation plan omits the existing zero-trip loop regression gate `docs/designs/ptoas-insertsync-memory-phi-canonical-writer-design.md`:`195`

Section 11.1 only gates `issue428` and `issue454`, but this ADR also rewrites loop-header/back-edge handling (`LoopPhi`, seed/drain, eventId decisions). The current implementation has dedicated zero-iteration correctness logic, and the repo already carries `test/lit/pto/issue533_loop_zero_trip_sync_regression.pto` to ensure post-loop waits are preserved when a loop executes zero times. Leaving that regression out of the must-pass list means a first-iteration/zero-trip regression in the new loop canonicalization can slip through CI.

PR #605, opened 2026-04-29
[ADR] VF fusion PIPE_V barrier coarsening design

TaoTao-real:
## Summary
Add an architecture-level ADR for reducing redundant `pipe_barrier(PIPE_V)` insertion in VF-fusible vector chains.

## Why
Current auto sync tends to insert PIPE_V barriers conservatively for same-pipe dependencies, which can over-serialize vector chains (e.g. FA/softmax style sequences) and leave performance on the table.

## What is included
- New design doc:
  - `docs/designs/ptoas-vf-fusion-pipev-barrier-optimization-adr.md`
- Proposed architecture:
  - VF segment modeling
  - risk-matrix based keep/drop decisions
  - unknown-safe conservative fallback
  - phased rollout and regression gates

## Scope
- Docs-only ADR PR
- No code/behavior change in this PR

## Follow-up
Implementation will be tracked in follow-up PRs with strict regression gates (issue428/454 + FA representative cases).

### DIFF
--- a/docs/designs/ptoas-vf-fusion-pipev-barrier-optimization-adr.md
+++ b/docs/designs/ptoas-vf-fusion-pipev-barrier-optimization-adr.md
@@ -1,0 +1,246 @@
+# ADR: VF 融合场景下 `PIPE_V` Barrier 精简优化
+
+- Status: Proposed
+- Date: 2026-04-29
+- Owners: PTOAS InsertSync maintainers
+- Scope: `InsertSync` 分析与生成层（`PIPE_V` 同 pipe barrier）
+
+---
+
+## 1. 背景
+
+当前 `InsertSync` 在同 pipe（`PIPE_V -> PIPE_V`）的同步策略较保守：  
+一旦识别到依赖，通常直接插入 `pipe_barrier(PIPE_V)`。  
+在 FA/softmax 等向量链路上，会出现“几乎每步一条 barrier”的过串行化现象。
+
+这与手工优化版本存在差异：手工版本常在“关键边界”插 barrier，而在可连续执行的 VF 链中不逐步插入。
+
+---
+
+## 2. 问题定义
+
+问题不是“完全移除 `PIPE_V` barrier”，而是区分：
+
+1. 必要 barrier：防止真实读写冲突、跨阶段可见性问题、特殊算子内部临时缓冲风险。
+2. 冗余 barrier：可由 VF 连续执行顺序天然保证，或可通过更粗粒度边界同步替代。
+
+目标是让自动同步在 VF 可融合链上“少插且不漏插”。
+
+---
+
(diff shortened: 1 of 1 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #605 [ADR] VF fusion PIPE_V barrier coarsening design
- Author: `TaoTao-real`
- Base/Head: `main` / `codex/adr-vf-pipev-barrier`
- Head SHA: `0a02c82cc99a`
- Trigger: 检测到新的 open PR
- Generated At: `2026-04-29T09:40:54Z`
- Status: completed

### Summary

The ADR's pairwise coarsening model can remove required same-pipe barriers, and its validation criteria are not aligned with the existing A5 `PIPE_V` barrier contract.

### Findings

1. **P1** Adjacent-pair VF segmentation can delete barriers that protect non-adjacent hazards `docs/designs/ptoas-vf-fusion-pipev-barrier-optimization-adr.md`:`112`

The proposal reasons about `PIPE_V` barriers only through adjacent `A -> B` pairs and segment boundaries. Current InsertSync is not local in that way: `InsertSeqSync` scans all earlier ops for each `now` op, and a same-pipe barrier before `now` records the original producer (`lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp`, around lines 153-176 and 350-356). A barrier before `C` can therefore represent an `A -> C` hazard even when the immediate predecessor `B` is unrelated. In a sequence like `A: write ub0; B: write ub1; C: read ub0`, both adjacent pairs look safe, but the barrier before `C` is still required. A segmenter that only decides on adjacent pairs would absorb `C` into the same segment and drop a required barrier, producing wrong results on same-pipe RAW/WAW chains. This also contradicts the current same-pipe contract exercised by `test/samples/Sync/test_inject_sync_intra_pipe_barrier.pto`.

2. **P2** The acceptance criteria ignore that A5 already strips \`PIPE_V\` intra-pipe barriers `docs/designs/ptoas-vf-fusion-pipev-barrier-optimization-adr.md`:`183`

The ADR treats dense `PIPE_V` barriers as a general runtime problem and uses a drop in `pipeVBarrierAfter` as a success criterion. That is not true for A5 today: `SyncCodegen::CreateBarrierOp` already returns early for `PIPE_V` barriers on A5, and `LoweringSyncToPipe` erases any remaining `PIPE_V` barriers there. So InsertSync-side coarsening does not change emitted A5 kernels, while on A2/A3 the same barrier is still the correctness mechanism for same-pipe hazards. Without scoping the target architecture and whether the metric is counted in syncIR versus generated code/runtime, this ADR can appear successful without affecting the kernels it is trying to speed up, or encourage applying an A5-specific assumption to architectures where the barrier is still required.

PR #612, opened 2026-05-01
refactor: enforce tilebuf-only pipeline

Zhendong404:


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #612 refactor: enforce tilebuf-only pipeline
- Author: `Zhendong404`
- Base/Head: `main` / `squash-tilebuf-planmem`
- Head SHA: `19bbcd3eca5f`
- Trigger: PR 有新提交
- Generated At: `2026-05-06T02:13:36Z`
- Previous Head SHA: `c3e1f7cffb2d`
- Status: completed

### Summary

PR breaks existing memref-based IR/tests and miscomputes subview alias offsets for non-row-major tiles.

### Findings

1. **P1** Memref-based IR is no longer accepted `include/PTO/IR/PTOOps.td`:`37`

This removes `AnyMemRef` from `PTODpsType` and the drivers also stop registering `MemRefDialect`. The repo still has a large lit corpus with `memref<...>` inputs (for example the `_gss.pto` companions under `test/lit/pto/`), so those files will now fail to parse/verify instead of round-tripping. That is a CI/regression blocker, not just a pipeline refactor.

2. **P2** Subview alias offsets now assume row-major layout `lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp`:`66`

`getStaticOffsetAndSize()` rebuilds strides from the logical shape and ignores the tile's actual layout/preserved stride. For `pto.subview` on col-major or non-compact tiles, the computed `baseAddresses` will be wrong, which feeds the sync/memory alias analysis and can suppress required waits or insert spurious ones.

PR #620, opened 2026-05-05
Add auto cv pipeline mulit buffer enable and preload support

zhangstevenunity:


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #620 Add auto cv pipeline mulit buffer enable and preload support
- Author: `zhangstevenunity`
- Base/Head: `main` / `codex/cv-preload-scope-marking`
- Head SHA: `ef2b6c6a3f71`
- Trigger: PR 有新提交
- Generated At: `2026-05-06T11:54:25Z`
- Previous Head SHA: `aa5ff4602e09`
- Status: completed

### Summary

PR #620 introduces two P1 correctness regressions in the new CV multi-buffer/preload pipeline and one P2 error-handling bug in PlanMemory.

### Findings

1. **P1** Default pipeline auto-marks CV multi-buffering without also enabling buffer-slot rotation `tools/ptoas/ptoas.cpp`:`1140`

`PTOCVAutoMarkMultiBufferPass` and `PTOCVMarkPreloadScopesPass` are now always added to the default pipeline, but `PTOEnableMultiBufferPass` still only runs when `--enable-multi-buffer-lowering` or `--enable-cv-create-preload` is set. On a plain `--enable-insert-sync` / `--enable-graph-sync-solver` build, PlanMemory can therefore emit variadic `pto.pointer_cast`s, the sync analyzers will use all N planned addresses to choose `eventIdNum > 1`, and SyncCodegen will emit `wait_flag_dyn` / `set_flag_dyn` rotation. EmitC later lowers the same `pto.pointer_cast` with `adaptor.getAddrs()[0]` only, so the runtime rotates event IDs but always reuses slot 0. That is a real miscompile for existing sync-only command lines on CV kernels, and it also changes memory planning even when the user never opted into multi-buffer/preload.

2. **P1** Create-preload leaves hoisted bind_tile-backed multi-buffer casts unlowered `lib/PTO/Transforms/PTOEnableMultiBuffer.cpp`:`146`

`PTOCVCreatePreloadPass` hoists cloned `pointer_cast` + `bind_tile` pairs above the rewritten loop. `inferRotationLoopFromUses()` only inspects the direct users of the `pointer_cast`; in the hoisted case that direct user is the hoisted `BindTileOp`, which is itself outside any `scf.for`, so the helper returns `nullptr` and the pass skips lowering entirely. The remaining variadic `pto.pointer_cast` then reaches EmitC, which again uses address 0 only. That means different preload stages alias the same physical local-buffer slot, breaking the advertised `--enable-cv-create-preload` path for normal tilebuf/bind_tile based kernels.

3. **P2** PlanMemory ignores multi-buffer offset-generation errors because the failure flag is checked too early `lib/PTO/Transforms/PTOPlanMemory.cpp`:`1528`

`emitMultiBufferError()` is meant to make the new retry loop reject a bad multi-buffer plan, but `plan()` checks `multiBufferDiagnosticEmitted_` before calling `UpdateBuffer2Offsets()`. The new invariants added in `UpdateBuffer2Offsets()` (`null relation slot`, `offset count mismatch`) therefore set the flag too late: `plan()` still returns success and the pass continues with a partially built `buffer2Offsets` map. That defeats the retry/fail-fast logic and can feed invalid addresses into `AllocToPointerCast` instead of stopping cleanly.

PR #651, opened 2026-05-11
Fix A5 GEMV verifier for aligned ACC rows

HecreReed:
## Summary
- stop reusing the A5 matmul static M/N shape check for GEMV-family ops
- keep the A5 left/right/acc layout+slayout enforcement shared between matmul and GEMV
- add an A5 lit regression covering `pto.tgemv`, `pto.tgemv.acc`, and `pto.tgemv.bias` with aligned ACC rows but `valid_shape[0] == 1`

## Validation
- `git diff --check`
- local macOS configure succeeded with `cmake -G Ninja -S . -B build-local ...`
- full local `ptoas` build is currently blocked by pre-existing macOS `-Wcovered-switch-default` errors in `lib/PTO/IR/PTOSyncUtils.cpp` and `lib/PTO/IR/PTOTypeDefs.cpp`

### DIFF
--- a/lib/PTO/IR/PTO.cpp
+++ b/lib/PTO/IR/PTO.cpp
@@ -137,6 +137,8 @@ static LogicalResult verifyMatTileOperandsA2A3(Operation *op, Type lhsTy,
                                                Type rhsTy, Type dstTy);
 static LogicalResult verifyMatTileOperandsA5(Operation *op, Type lhsTy,
                                              Type rhsTy, Type dstTy);
+static LogicalResult verifyMadTileLayoutsA5(Operation *op, Type lhsTy,
+                                            Type rhsTy, Type dstTy);
 static LogicalResult verifyGemvTileOperands(Operation *op, Type lhsTy, Type rhsTy,
                                             Type dstTy);
 static LogicalResult verifyGemvTileOperandsA2A3(Operation *op, Type lhsTy,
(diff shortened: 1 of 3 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #651 Fix A5 GEMV verifier for aligned ACC rows
- Author: `HecreReed`
- Base/Head: `main` / `codex/a5-gemv-verifier-fix`
- Head SHA: `2b186ac481ee`
- Trigger: PR 有新提交
- Generated At: `2026-05-11T02:23:45Z`
- Previous Head SHA: `2ffb991312fe`
- Status: completed

### Summary

A5 GEMV verifier regressed in this PR: the new helper call fixes aligned ACC rows, but it also drops prior range and logical-shape validation for all A5 GEMV variants.

### Findings

1. **P2** A5 GEMV no longer rejects out-of-range valid sizes `lib/PTO/IR/PTO.cpp`:`3214`

`verifyGemvTileOperandsA5()` used to delegate to `verifyMatTileOperandsA5()`, which in turn enforced the `m/k/n` valid-size bounds from `verifyMatTileOperandsA2A3()` (`[1, 4095]`). After this change it returns `verifyMadTileLayoutsA5(...)` directly, so every A5 GEMV variant now loses that verifier coverage. That means `pto.tgemv`, `pto.tgemv.acc`, `pto.tgemv.bias`, and the `tgemv.mx*` ops can now accept `k`/`n` values above 4095 that were previously rejected and are still documented as invalid, which is a real verifier regression with runtime-risky inputs.

2. **P2** A5 GEMV shape compatibility checks were over-relaxed `lib/PTO/IR/PTO.cpp`:`3214`

This refactor fixes aligned ACC rows by bypassing `verifyMatTileOperandsA5()`, but that call was also the only A5 GEMV path that checked full logical `lhs[M,K] / rhs[K,N] / dst[M,N]` compatibility through `getMatmulLogicalShapeVec()`. The replacement only checks a subset of `valid_shape` relationships, so malformed A5 GEMV descriptors can now slip through whenever one side uses dynamic or padded extents. For example, a `dst` tile with a mismatched static column shape but dynamic `v_col` now passes verification even though the GEMV contract still requires `rhs.cols == dst.cols`. That widens the accepted IR beyond the intended "aligned ACC rows" relaxation and can reach codegen/runtime with inconsistent tile metadata.

HecreReed:
Closing without merge; GEMV verifier work will be re-proposed with current main as needed.

PR #685, opened 2026-05-19
Add explicit multi-buffer annotation and manual buffer select support

zhangstevenunity:


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #685 Add explicit multi-buffer annotation and manual buffer select support
- Author: `zhangstevenunity`
- Base/Head: `main` / `claude/serene-bhaskara-f0221e`
- Head SHA: `fd3f844c965e`
- Trigger: PR 有新提交
- Generated At: `2026-06-30T10:29:03Z`
- Previous Head SHA: `e0e56ee26aa4`
- Status: completed

### Summary

PR #685 has two correctness issues: level3 explicit-address multi-buffers lose their multi-buffer sync semantics in InsertSync, and multi_tile_buf values crossing function boundaries are silently lowered as plain single-buffer memrefs.

### Findings

1. **P1** InsertSync collapses level3 explicit-address multi-buffers to a single event lane `lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp`:`564`

`pto.alloc_multi_tile addr=...` in level3 is lowered to a multi-address `pto.pointer_cast`, but slot 1..N-1 addresses are always built with `arith.addi` in `PTOViewToMemref`. `UpdatePointerCastOpMemInfo` only keeps multi-buffer geometry when every address operand is a direct `arith.constant`; otherwise it hits the fallback at this line and records the cast as a single-buffer root with `baseAddresses={0}`. That means every downstream `pto.slot_marker` loses its N-slot address set, `GetEventIdNum()` falls back to 1, and InsertSync emits one static set/wait instead of per-slot dyn event IDs for back-edge hazards. In the alternating-slot prefetch patterns this PR adds, that can let a wait on slot k be satisfied by a set from slot k' under `--pto-level=level3 --enable-insert-sync`, i.e. wrong synchronization rather than just lost optimization. This happens even when the base addr is constant, because slot 1..N-1 are still `arith.addi` results.

2. **P2** Unsupported multi_tile_buf function signatures are accepted and silently downgraded `lib/PTO/Transforms/PTOViewToMemref.cpp`:`764`

The design and in-code comments say `!pto.multi_tile_buf` must not appear on function arguments or results, but the lowering here silently rewrites such types to the underlying single-slot memref shape instead of rejecting them. Because function/block arguments do not carry the `pto.multi_buffer` attribute, the slot count is lost at the boundary. A later `pto.multi_tile_get` on that value becomes a `pto.slot_marker` with no root multi-address `pto.pointer_cast`, so `PTOResolveBufferSelect` just forwards the source memref and all slots alias the same buffer. In other words, a helper function that takes or returns a multi-buffer is accepted and compiled as single-buffer code with no diagnostic. This is a real contract mismatch and wrong-code path, not just an unsupported feature being rejected.

tools/ptoas/ptoas.cpp
```diff
@@ -1901,6 +1904,11 @@ int main(int argc, char **argv) {
         pto::createPTOGraphSyncSolverPass(graphSyncOpts));
   }
 
+  // Materialize per-slot single-address `pto.pointer_cast` (constant slot)
+  // or an `arith.select` chain (dynamic slot). The multi-address cast
+  // produced by PlanMemory survives as the alloc anchor.
+  pm.addPass(pto::createPTOResolveBufferSelectPass());
```
zhangstevenunity, line 1918: **P1 (confirms Codex finding 1): level3 silently downgrades multi-buffer to single-buffer.**

This pass runs unconditionally, but PlanMemory -- which expands the `pto.multi_buffer=N` `memref.alloc` into the multi-address `pto.pointer_cast` anchor -- is skipped when `--pto-level=level3` (ptoas.cpp:1875). With no multi-address anchor, `lookupRootPointerCast` in PTOResolveBufferSelect returns null and the `!root` branch replaces every `pto.slot_marker` with the original single memref, so `%mb[0]`, `%mb[1]`, ... all alias the same physical storage -- a silent wrong result with no diagnostic. The level3 entry block (ptoas.cpp:1800) only walks `AllocTileOp` to require `addr`; it does not reject multi-buffer ops. Suggest rejecting `AllocMultiTileOp` / `MultiTileGetOp` under level3 there with an actionable error.

lib/PTO/Transforms/PTOViewToMemref.cpp
```diff
@@ -597,10 +605,16 @@ static Type convertPTOTypeToMemRef(Type t) {
...
-  
+
   // 2. 处理 !pto.tile_buf<...>
   if (auto tbTy = dyn_cast<mlir::pto::TileBufType>(t))
     return convertTileBufTypeToMemRef(tbTy);
+  // 3. !pto.multi_tile_buf<S, count=N>: collapses to the slot memref shape;
+  //    the N-slot fan-out lives on the `pto.multi_buffer` attr written by
+  //    the alloc lowering, not in the type. This branch is defensive --
+  //    by design multi_tile_buf does not appear on function boundaries.
+  if (auto mtbTy = dyn_cast<mlir::pto::MultiTileBufType>(t))
```
zhangstevenunity, line 764: **P2 (confirms Codex finding 2): a function-boundary multi_tile_buf is silently collapsed; the documented restriction is unenforced.**

This branch (and the function-signature rewrite at Stage 0.10, ~line 1406) unconditionally rewrites `!pto.multi_tile_buf<S,count=N>` to the single slot memref, dropping N and the `pto.multi_buffer` attr. The design doc (docs/designs/ptoas-multi-buffer-explicit-design.md:253,470) and PTO_IR_manual.md:191 say multi_tile_buf is not allowed on func args/returns -- the manual even lists it under "Constraints (enforced by the type verifier)" -- but nothing enforces it (MultiTileBufType::verify, AllocMultiTileOp::verify, MultiTileGetOp::verify do no boundary check). If such IR ever reaches here the N slots silently become 1 => reintroduced WAR/WAW race. Add a real verifier rejecting MultiTileBufType on func.func arg/result types, or at minimum correct the manual to say "frontend contract, not verified."

lib/PTO/Transforms/InsertSync/SyncCodegen.cpp
```diff
@@ -350,15 +350,58 @@ void SyncCodegen::CreateSetWaitOpForMultiBuffer(IRRewriter &rewriter,
...
+    auto eventId = getEventAttr(rewriter, sync->eventIds[0]);
+    createSetOrWaitFlagOp(rewriter, op, sync, srcPipe, dstPipe, eventId);
+    return;
+  }
+
+  // Emit pto.set_flag_dyn / pto.wait_flag_dyn with a runtime event id chosen
+  // from the slot SSA. Specifically, build an N-way select chain over the
+  // allocated event ids so the hardware sees event id `eventIds[slot % N]`.
+  // For the common N == 2 case this collapses to a single arith.select.
+  uint32_t n = sync->slotCount;
```
zhangstevenunity, line 373: **CORRECTION (2026-06-30) -- retracting the P1 below: it is NOT reachable. Not an OOB / data race.**

I tried to build a triggering IR: a multi-buffer `count=9..16` prefetch loop, which forces event-id exhaustion because only `kTotalEventIdNum = 8` exist per pipe pair. Ran it through `--enable-insert-sync`. It does NOT read out of bounds; it degrades safely to a single static event id.

Why the realloc case is safe: the fallback (SyncEventIdAllocation.cpp:163-168) reduces a starved multi-buffer sync to `eventIds.size() == 1`. The codegen dispatch in `SyncInsert` (SyncCodegen.cpp:275) routes `eventIds.size() == 1` to `CreateSetWaitOpForSingleBuffer` -- so `CreateSetWaitOpForMultiBuffer` (and its `for i in 1..slotCount-1 { eventIds[i] }` loop here) is never entered for the collapsed sync. `slotCount` staying N is simply ignored by the single-buffer path.

More generally, `SetEventId` allocates all-or-nothing: a multi-buffer sync ends up with `eventIds.size()` == N (normal, matches slotCount), == 1 (realloc -> single-buffer dispatch), or 0/useless (widen-covered, or PipeAll barrier via `ChangeNoEventIdSyncToPipeAll`). It is never in `[2, slotCount-1]`. So `CreateSetWaitOpForMultiBuffer` is only ever reached with `eventIds.size() == slotCount`, and the loop + the `assert(eventIds.size() == n)` at line 375 always hold. Verified empirically (asserts compiled in): count=2 -> dyn (2 events); count=9 / count=16 -> static single event id, no assert, no crash.

Residual (minor, NOT a blocker): a multi-buffer with `count > kTotalEventIdNum` (8) can never obtain N dyn event ids, so it silently loses per-slot pipelining and serializes through one event id (conservative over-synchronization -- correct, just slower). A verifier note/warning that `count > 8` is not fully pipelinable under InsertSync would be a reasonable follow-up. Apologies for the false alarm on the OOB.

lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp
```diff
@@ -740,16 +809,40 @@ SmallVector<Value> InsertSyncAnalysis::GetMemInfoBuffers(
...
     bool isLocalA =
         pair.first && (pair.first->scope == pto::AddressSpace::MAT ||
                        pair.first->scope == pto::AddressSpace::VEC);
     bool isLocalB =
         pair.second && (pair.second->scope == pto::AddressSpace::MAT ||
                         pair.second->scope == pto::AddressSpace::VEC);
-    if (isLocalA || isLocalB) return 1;
+    if (!isLocalA && !isLocalB)
+      continue;
+    size_t aN = pair.first ? pair.first->baseAddresses.size() : 1;
```
zhangstevenunity, line 831: **CORRECTION (2026-06-30): downgrading P2 -> P3. This is NOT an under-sync. The premise holds, but the consequence does not.**

Verified premise: `getPtoSubViewBaseAddresses` (PTOIRTranslator.cpp:131-136) returns `rowSize` addresses for ANY row-major tile subview, so `GetEventIdNum` does return > 1 for an ordinary subview, and the `MemAnalyze` lane-skip loop (line 496-503) calls it without the slot-marker guard `InsertSyncOperation` applies (line 637-642). That asymmetry is real.

But I could not turn it into a dropped sync, for two reasons:

1. **Back-edge analysis uses a per-compound FRESH `SyncRecordList`.** `InsertBackForSync` declares `SyncRecordList syncRecordList;` (line 263) and is called once per compound (`DealWithLoopSync` line 244). So a multi-buffer compound's lane marks (lanes 1..N-1, lane 0 skipped) and a separate subview compound's lane-1..K-1 check live in DIFFERENT `SyncRecordList`s -- the cross-compound lane pollution my original comment described cannot happen. (The forward/same-iteration path only ever checks lane 0, line 468.)

2. **`GetEventIdNum` returns 1 when dep pairs disagree on N** (line 838-842). A multi-buffer pair (N = slot count) and a subview pair (N = rowSize) disagree, so the lane loop does not run for a mixed compound. The asymmetry-creating multi-buffer sync (eventIdNum>1, skips lane 0) and a subview check that runs the lane loop therefore cannot coexist in one fresh `SyncRecordList`.

Empirically confirmed: a loop with a `count=2` multi-buffer prefetch (MTE3->MTE2 dyn) PLUS an independent 16x16 subview back-edge WAR on the same MTE3 pipe keeps the subview's `wait_flag[MTE3, MTE2, EVENT_ID2]` -- the subview WAR is correctly synced, not dropped.

Residual (P3 cleanup, not a blocker): `GetEventIdNum` deriving N from `baseAddresses.size()` (which counts subview row segments, not just multi-buffer slots) is fragile -- it stays correct only because of the two guards above. Folding the slot-marker check (`findSlotMarkerExpr` non-null) into `GetEventIdNum`, so "N" means "multi-buffer slot count" at every call site, would make the invariant local and robust against future refactors. Apologies -- my original P2 mechanism was wrong.

zhangstevenunity (commented):
## Multi-buffer PR review (head `e0e56ee2`)

Note: I am the PR author on GitHub, so this is posted as a COMMENT review rather than a formal request-changes (GitHub blocks self request-changes). Treat the P1/P2 items below as blocking-quality findings.

### Mergeability -- hard blocker
This branch is CONFLICTING / DIRTY against `main` and has not been updated since 2026-06-02 (28 days). `main` has moved ~2756 files since the merge-base (`40505dca`), and 15 of this PR's 49 files were also changed on `main`, including the exact subsystems this PR rewires: PTOPlanMemory, PTOViewToMemref, InsertSync/PTOIRTranslator, GraphSyncSolver/SyncSolverIRTranslator, ptoas.cpp, PTOOps.td, PTOTypeDefs.td. It cannot land without a non-trivial rebase, after which the sync integration must be re-validated.

### What is solid
- Build is green at head (the earlier TableGen missing-brace break is fixed; CI build-and-test + both arch builds + vpto-sim-validation pass).
- The GSS dyn event-id slot-binding (commits `f3729c06` + `e0e56ee2`) is sound: the set side keys off the producer's write-slot and the wait side off the consumer's read-slot of the hazard buffer; ambiguous/absent slot falls back to the safe symmetric N-static fanout (re-checked against the prefetch and two-buffer regression tests).
- PlanMemory forces sibling slots disjoint (shared lifetime => no reuse-overlap; the post-plan overlap guard catches violations).
- MultiTileBufType parse/print round-trips; the `Pure` traits on multi_tile_get / slot_marker are CSE-safe (distinct slot operands are never merged).

### Findings (see inline comments)
- **P1** `--pto-level=level3`: multi-buffer is silently downgraded to single-buffer (all slots alias). [Codex finding 1, confirmed]
- **P1** InsertSync event-id exhaustion vs. the new `slotCount` field: OOB read / wrong event lane in release builds.
- **P2** `multi_tile_buf` on a function boundary is silently collapsed to one slot with no diagnostic (the manual claims this is "enforced by the type verifier" -- it is not). [Codex finding 2, confirmed]
- **P2** `GetEventIdNum` derives N from `baseAddresses.size()`, which is also >1 for ordinary strided subviews; the MemAnalyze early-skip loop consumes it without the slot-marker guard that InsertSyncOperation applies.
- **P3** `alloc_multi_tile` rejects low-precision dtypes that `alloc_tile` accepts (safe conservative guard, but inconsistent with the op and the manual). [Codex finding 3, confirmed -- downgraded from P2]
- **P3** `compareSlotSSA` callers do not confirm both slot markers root to the same multi-buffer alloc (latent; cheap to harden).

### Minor / follow-up
- SlotAffineAnalysis: the pure-constant modulus-mismatch sub-case projects onto N not the form modulus M (SlotAffineAnalysis.cpp:181); int64 offset overflow across 4 peels is theoretical (130).
- `getOperationAliasInfo` now traces through `pto.bind_tile` for ALL files in the GSS path (Utils.cpp:88), not just multi-buffer -- likely a correctness improvement, but there is no before/after GSS regression test proving equivalence on existing bind_tile programs.
- Python ergonomic API + test/samples are deferred by design (generated op bindings still build, so no hard break); add a Python smoke test to lock the auto-generation path if the cross-layer rule is read strictly.

lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp
```diff
@@ -436,6 +438,29 @@ void InsertSyncAnalysis::InsertSync(
...
+static bool isForwardDepDroppableBySlotAffine(const BaseMemInfo *a,
+                                              const BaseMemInfo *b) {
+  if (!a || !b)
+    return false;
+  size_t aN = a->baseAddresses.size();
+  size_t bN = b->baseAddresses.size();
+  size_t n = std::max(aN, bN);
+  if (n < 2)
+    return false;
+  Value slotA = findSlotMarkerExpr(a->baseBuffer);
```
zhangstevenunity, line 456: **P3 (latent / defensive): slot-disjointness is asserted without confirming both markers index the same multi-buffer.**

`isForwardDepDroppableBySlotAffine` (here) and the GSS twin `compareMemInfoSlotSSA` (SyncSolver.cpp:51-55) compare two slot SSA exprs purely arithmetically; `(i+1)%2 != i%2` proves disjointness only when both markers index ONE shared N-address fan-out. Neither caller checks that `slotA` and `slotB` derive from the same alloc anchor. This pass is only reached because the alias analysis already found an address overlap; if two different multi-buffers ever shared storage, a real forward dep between them would be dropped. I could not construct a concrete trigger (PlanMemory only reuses storage across disjoint lifetimes, which fights a live forward dep), so this is latent -- but cheap to harden: return kUnknown unless both markers walk back to the same root pointer_cast / alloc.

lib/PTO/IR/PTO.cpp
```diff
@@ -2476,6 +2477,87 @@ LogicalResult AllocTileOp::verify() {
...
+  if (!mtbTy)
+    return emitOpError("result must be `!pto.multi_tile_buf`");
+
+  TileBufType slotTy = mtbTy.getSlotType();
+  if (!slotTy)
+    return emitOpError("multi_tile_buf slot type must be non-null");
+
+  // Reuse the AllocTileOp valid_row/valid_col contract on the slot type.
+  Type elemTy = slotTy.getElementType();
+  if (isPTOLowPrecisionType(elemTy))
```
zhangstevenunity, line 2867: **P3 (confirms Codex finding 3, downgraded from P2): conservative low-precision guard -- safe but inconsistent.**

This rejects low-precision slot dtypes (fp8 / hif8 / fp4) that plain `AllocTileOp::verify` (PTO.cpp:2443) accepts, and PTO_IR_manual.md:767 says the slot is "verified the same way as alloc_tile." Tracing the multi-buffer lowering (PTOViewToMemref reuses the alloc_tile stride/layout machinery; PlanMemory sizes via getPTOStorageElemByteSize which already maps these types to byte sizes; ResolveBufferSelect does no element math), nothing would miscompile a low-precision slot if the guard were removed. So this is a safe "not supported yet" limitation, not a correctness bug: keep it, but add a TODO/tracking-issue reference and a manual bullet, and lift it once a multi-buffer low-precision lit test exists. (Same point as the inline Gemini note.)

reedhecre:
已接收 `/run all`，A3 板测器会处理这条请求。

- 进度页：http://154.9.227.233/ptoas-board-dashboard/
- 当前状态：板测器空闲，这条请求会在本轮轮询启动。

页面会自动刷新，可以直接看当前阶段、排队情况和最近结果。

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`manual`
- 源码提交：`13ef90a9b6c5`
- 结果汇总：OK `221` / FAIL `0` / SKIP `1`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260630_181309_manual_pr685.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260630_181309_manual_pr685.tsv`
- 手动指令：`/run all`
- 触发人：`zhangstevenunity`
- 触发评论：https://github.com/hw-native-sys/PTOAS/pull/685#issuecomment-4842040796

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`28f256dd0e5d`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_111508_merged_pr685.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_111508_merged_pr685.tsv`

PR #731, opened 2026-06-01
Several cube DSL related bugfixes

Zhendong404:
## Summary
This PR cherry-picks the 8 commits in the inclusive range [0443c37e3, 51f322407] onto the latest main.

Included fixes:
- Declare TileLang daemon POSIX symbols
- Fix ExpandTileOp memory space stringify
- Support constraints in ckernel
- Convert positional dtype strings to real dtypes in DSL lowering
- Support `si32` pre_quant type in DSL
- Support signless `i8*i8->i32` mad lowering on VPTO
- Fix daemon core scalar value passing in DSL
- Restrict `indexRow`/`indexCol` to `0` for `textract` mat -> left/right

## Validation
- `PYTHONPATH=$PWD/tilelang-dsl/python python3 -m unittest tilelang-dsl/tests/test_tilelang_dsl_v1.py`

## Notes
- I attempted to run the related `lit` tests as well, but the available generated `lit.site.cfg.py` in this environment points at a different source checkout, so I did not count that as a reliable validation result for this isolated PR branch.

### DIFF
--- a/lib/PTO/IR/PTO.cpp
+++ b/lib/PTO/IR/PTO.cpp
@@ -5735,6 +5735,18 @@ mlir::LogicalResult mlir::pto::TExtractOp::verify() {
     if (*srcSpace == pto::AddressSpace::MAT) {
       if (!hasMatExtractSourceLayoutA5(srcTb, *dstSpace))
         return emitOpError("expects A5 textract src to use a supported mat blayout/slayout combination");
+      if (*dstSpace == pto::AddressSpace::LEFT ||
+          *dstSpace == pto::AddressSpace::RIGHT) {
+        auto indexRowVal = getConstantIntegerValueEx(
+            getIndexRow(), /*includeIndexAndIntOpsInConstFold=*/true);
+        auto indexColVal = getConstantIntegerValueEx(
+            getIndexCol(), /*includeIndexAndIntOpsInConstFold=*/true);
+        if (!indexRowVal || !indexColVal || *indexRowVal != 0 ||
+            *indexColVal != 0)
+          return emitOpError(
+              "expects A5 mat->left/right textract to have indexRow=0 and "
+              "indexCol=0 (offset extraction not yet supported)");
+      }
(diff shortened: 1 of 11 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #731 Serveral bugfixs
- Author: `Zhendong404`
- Base/Head: `main` / `codex/several-bugfixs`
- Head SHA: `8269e3f5ae9c`
- Trigger: 检测到新的 open PR
- Generated At: `2026-06-01T04:29:10Z`
- Status: completed

### Summary

PR #731 introduces a hard A5 `textract` regression that breaks existing sample/CI coverage, and the new daemon scalar-constraint support is not safe with the current cache key.

### Findings

1. **P1** A5 \`textract\` zero-offset restriction breaks existing decode samples and CI `lib/PTO/IR/PTO.cpp`:`5744`

The new A5 verifier rejects any `pto.textract` from `loc=mat` to `loc=left/right` unless both indices fold to zero. That immediately breaks existing checked-in A5 level3 samples that use non-zero half-tile offsets, for example `test/samples/Qwen3DecodeA5/qwen3_decode_incore_1.pto` lines 48-50 (`%c64_index`). `test/samples/runop.sh` compiles those direct `.pto` samples, and `.github/workflows/ci.yml` runs that path in CI, so this change turns current sample coverage into a hard compile failure.

2. **P2** Instance cache does not key on scalar values used by new daemon constraints `tilelang-dsl/python/tilelang_dsl/stable_key.py`:`338`

This PR adds daemon-side constraint evaluation over scalar constants via `argN_value`, but the cache key still hashes scalar operands only by kind and dtype. In a long-lived `InstanceCache`, two requests that differ only by a scalar value will alias to the same key, so after the first instantiation the daemon can return MLIR for the wrong descriptor on later requests. The new test avoids this by creating a fresh cache per call, so the production cache path is still broken.

Zhendong404:
Already merged in another PR

PR #753, opened 2026-06-04
feat(pto): add !pto.struct heterogeneous aggregate type and ops

zhangstevenunity:
Add a scalar-unit heterogeneous aggregate !pto.struct as the heterogeneous dual of !pto.local_array, mirroring its end-to-end pattern with minimal new mechanism.

- ODS: !pto.struct<T0,...> type + declare_struct / struct_get / struct_set (constant DenseI64ArrayAttr field path, custom<StructPath> assembly).
- IR: type verifier restricts fields to scalar int/float, nested local_array, or nested struct (rejects tile_buf/tensor_view and other handle types, keeping the scalar world disjoint from the vec/cube world); op verifiers walk the path with bounds/type checks and reject whole-array struct_set.
- EmitC: StructType -> emitc.opaque<PtoStruct_<mangled>>; file-scope struct definitions emitted in dependency order (nested first); member-access chain lowering with scalar snapshot semantics, while aggregate fields pass through as lvalues so a nested local_array field can be indexed via local_array_get/set.
- Docs (PTO_IR_manual section 2.7 + op references) and 5 lit tests (scalar/nested codegen + 3 verifier-error cases).

### DIFF
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -306,6 +306,22 @@ the scalar struct world separate from the fractal/layout world, exactly like
   // { PtoStruct_f32_i8 f0; int16_t f1; };
 ```
 
+**Passing a struct to a function.** A `!pto.struct` value lowers to a **pointer**
+to its storage, so it can be passed to a helper `func.func` and mutated there:
+
+```mlir
+func.func private @helper(%s : !pto.struct<f32, i8>, %v : f32) {
+  pto.struct_set %s[0], %v : !pto.struct<f32, i8>, f32   // emits `p->f0 = v;`
+  return
+}
+```
+
+The callee reaches fields through `->`; the caller declares real storage, passes
+its address, and still uses `.` on its own copy. Carrying the struct by value
+would make the callee's writes land in a copy, and an `!emitc.lvalue` is
+rejected outright as a function argument type by both `emitc.func` and the C++
+emitter — the pointer is what makes helper functions expressible at all.
+
 **Generated type name.** The C++ name is derived from the MLIR type spelling of
 the fields (`!pto.struct<f16, i8>` → `PtoStruct_f16_i8`), not from their C++
 spellings. That mapping is injective, so two distinct `!pto.struct` types never
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -274,19 +274,25 @@ positionally numbered `f0, f1, ...`, matching an LLVM literal struct.
 
 | Parameter | Type | Constraints |
 |-----------|------|-------------|
-| `fieldTypes` (`T0..Tn-1`) | `i8`/`i16`/`i32`/`i64` (signed, unsigned or signless), `f16`/`bf16`/`f32`/`f64`, a `!pto.local_array` of those, or a nested `!pto.struct` | At least one field; recursively scalar-storable |
+| `fieldTypes` (`T0..Tn-1`) | `i8`/`i16`/`i32`/`i64` (signed, unsigned or signless), `f16`/`bf16`/`f32`/`f64`, or a nested `!pto.struct` | At least one field; recursively scalar-storable |
 
 **Constraints (enforced by the type verifier):**
 
 - At least one field
 - Each field is *scalar-storable*: an integer of width 8/16/32/64, one of
-  `f16`/`bf16`/`f32`/`f64`, a nested `!pto.local_array` whose element type is
-  one of those, or a nested `!pto.struct`
+  `f16`/`bf16`/`f32`/`f64`, or a nested `!pto.struct`
 
 A scalar field must map onto a C++ scalar the backend can name exactly. Integer
 widths with no C++ spelling (`i1`, `i24`, ...) and the packed low-precision
 vec/cube formats (`f8`/`f4` variants) are therefore **rejected**: emitting them
 would silently widen or narrow the field in the generated C++.
+
+**`!pto.local_array` is not a valid field type.** A field is reached with
+`emitc.member`, whose result must be an `!emitc.lvalue`, and `!emitc.lvalue`
+cannot wrap the `!emitc.array` that an array field lowers to. There is no way to
+spell the access, so the type verifier rejects it rather than letting the
+backend fail later. Use a nested `!pto.struct` of scalars, or keep the array as
+a separate `!pto.declare_local_array` value alongside the struct.
 
 **Disjoint from tile-buf world.** Vec/cube types (`tile_buf`, `tensor_view`,
 partition view) and other handle types are **rejected** as fields. This keeps
@@ -295,9 +301,9 @@ the scalar struct world separate from the fractal/layout world, exactly like
 
 **Syntax:**
 ```mlir
-!pto.struct<f16, i8>                                 // { half f0; int8_t f1; };
-!pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>>
-  // { PtoStruct_f32_i8 f0; float f1[4]; };
+!pto.struct<f16, i8>                          // { half f0; int8_t f1; };
+!pto.struct<!pto.struct<f32, i8>, i16>
+  // { PtoStruct_f32_i8 f0; int16_t f1; };
 ```
 
 **Generated type name.** The C++ name is derived from the MLIR type spelling of
@@ -313,10 +319,12 @@ collide on one name — mangling from the C++ token instead would give `i32` and
 
 Field positions in `get` / `set` are a constant `path` of indices that may
 descend through nested structs (like LLVM `extractvalue` / `insertvalue`).
-**Dynamic** indexing of a nested `!pto.local_array` field is *not* part of the
-path: obtain the array with `pto.struct_get`, then index it with
-`pto.local_array_get` / `pto.local_array_set` — the same split LLVM makes
-between constant struct GEPs and dynamic array indexing.
+
+The path must **end at a scalar**. A path that stops on a nested `!pto.struct`
+is rejected: the member chain lowers to `emitc.member`, which yields an lvalue,
+and returning a whole aggregate as an SSA value would copy it out of the parent
+struct — so a field inside a nested struct is reached with a longer path
+(`%s[0, 1]`) rather than by first getting the inner struct.
 
 ---
 
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -274,13 +274,19 @@ positionally numbered `f0, f1, ...`, matching an LLVM literal struct.
 
 | Parameter | Type | Constraints |
 |-----------|------|-------------|
-| `fieldTypes` (`T0..Tn-1`) | scalar `int`/`float`, `!pto.local_array`, or nested `!pto.struct` | At least one field; recursively scalar-storable |
+| `fieldTypes` (`T0..Tn-1`) | `i8`/`i16`/`i32`/`i64` (signed, unsigned or signless), `f16`/`bf16`/`f32`/`f64`, a `!pto.local_array` of those, or a nested `!pto.struct` | At least one field; recursively scalar-storable |
 
 **Constraints (enforced by the type verifier):**
 
 - At least one field
-- Each field is *scalar-storable*: a scalar integer/float, a nested
-  `!pto.local_array`, or a nested `!pto.struct`
+- Each field is *scalar-storable*: an integer of width 8/16/32/64, one of
+  `f16`/`bf16`/`f32`/`f64`, a nested `!pto.local_array` whose element type is
+  one of those, or a nested `!pto.struct`
+
+A scalar field must map onto a C++ scalar the backend can name exactly. Integer
+widths with no C++ spelling (`i1`, `i24`, ...) and the packed low-precision
+vec/cube formats (`f8`/`f4` variants) are therefore **rejected**: emitting them
+would silently widen or narrow the field in the generated C++.
 
 **Disjoint from tile-buf world.** Vec/cube types (`tile_buf`, `tensor_view`,
 partition view) and other handle types are **rejected** as fields. This keeps
@@ -291,8 +297,14 @@ the scalar struct world separate from the fractal/layout world, exactly like
 ```mlir
 !pto.struct<f16, i8>                                 // { half f0; int8_t f1; };
 !pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>>
-  // { PtoStruct_float_int8_t f0; float f1[4]; };
+  // { PtoStruct_f32_i8 f0; float f1[4]; };
 ```
+
+**Generated type name.** The C++ name is derived from the MLIR type spelling of
+the fields (`!pto.struct<f16, i8>` → `PtoStruct_f16_i8`), not from their C++
+spellings. That mapping is injective, so two distinct `!pto.struct` types never
+collide on one name — mangling from the C++ token instead would give `i32` and
+`si32` (both `int32_t`) the same name and emit a duplicate `struct` definition.
 
 **Associated ops** (see Section 4 — mirrors the `local_array` triad):
 - `pto.declare_struct` — declare
@@ -10471,7 +10483,7 @@ Section 2.7 for type-level constraints.
 **Basic Example:**
 
 ```mlir
-%s = pto.declare_struct -> !pto.struct<f16, i8>   // PtoStruct_half_int8_t s;
+%s = pto.declare_struct -> !pto.struct<f16, i8>   // PtoStruct_f16_i8 s;
 ```
 
 ---
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -264,6 +264,50 @@ The number of indices on `get` / `set` must equal the array's rank
 
 ---
 
+### 2.7 `!pto.struct<T0, T1, ..., Tn-1>`
+
+A **C++ stack-local heterogeneous aggregate** — the heterogeneous dual of
+`!pto.local_array`. Lowers to a plain `struct PtoStruct_X { ... };` definition
+plus a stack variable. The struct's address is decided by the host C++
+compiler, never by PTO memory planning or `pto.pointer_cast`. Fields are
+positionally numbered `f0, f1, ...`, matching an LLVM literal struct.
+
+| Parameter | Type | Constraints |
+|-----------|------|-------------|
+| `fieldTypes` (`T0..Tn-1`) | scalar `int`/`float`, `!pto.local_array`, or nested `!pto.struct` | At least one field; recursively scalar-storable |
+
+**Constraints (enforced by the type verifier):**
+
+- At least one field
+- Each field is *scalar-storable*: a scalar integer/float, a nested
+  `!pto.local_array`, or a nested `!pto.struct`
+
+**Disjoint from tile-buf world.** Vec/cube types (`tile_buf`, `tensor_view`,
+partition view) and other handle types are **rejected** as fields. This keeps
+the scalar struct world separate from the fractal/layout world, exactly like
+`!pto.local_array`.
+
+**Syntax:**
+```mlir
+!pto.struct<f16, i8>                                 // { half f0; int8_t f1; };
+!pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>>
+  // { PtoStruct_float_int8_t f0; float f1[4]; };
+```
+
+**Associated ops** (see Section 4 — mirrors the `local_array` triad):
+- `pto.declare_struct` — declare
+- `pto.struct_get`     — `s.fA.fB...` field read by constant path
+- `pto.struct_set`     — `s.fA.fB... = v;`
+
+Field positions in `get` / `set` are a constant `path` of indices that may
+descend through nested structs (like LLVM `extractvalue` / `insertvalue`).
+**Dynamic** indexing of a nested `!pto.local_array` field is *not* part of the
+path: obtain the array with `pto.struct_get`, then index it with
+`pto.local_array_get` / `pto.local_array_set` — the same split LLVM makes
+between constant struct GEPs and dynamic array indexing.
+
+---
+
 ## 3. Enums & Attributes
 
 ### 3.1 AddressSpace
@@ -10401,6 +10445,102 @@ Operates on the [`!pto.local_array<...>`](#26-ptolocal_arrayd1-x-d2-x--x-dk-x-t)
 ```mlir
 pto.local_array_set %m[%i, %j], %v
    : !pto.local_array<4x8xf32>, f32                        // m[i][j] = v;
+```
+
+---
+
+#### Struct Ops (`pto.declare_struct` / `pto.struct_get` / `pto.struct_set`)
+
+C++ stack-local heterogeneous aggregate ops. Disjoint from the tile-buf /
+scratch memory world: the struct's address is decided by the host C++
+compiler. Naming and asm style mirror the `local_array` triad.
+
+Operates on the [`!pto.struct<...>`](#27-ptostructt0-t1--tn-1) type. See
+Section 2.7 for type-level constraints.
+
+##### `pto.declare_struct` - Declare a Stack-Local Struct
+
+**Summary:** Declare a heterogeneous aggregate on the C++ stack.
+
+**Semantics:** `PtoStruct_X s;`
+
+**Arguments:** None.
+
+**Results:** `!pto.struct<T0, ..., Tn-1>`
+
+**Basic Example:**
+
+```mlir
+%s = pto.declare_struct -> !pto.struct<f16, i8>   // PtoStruct_half_int8_t s;
+```
+
+---
+
+##### `pto.struct_get` - Read a Field by Constant Path
+
+**Summary:** Read the field reached by a constant `path` of field indices,
+descending through nested structs.
+
+**Semantics:** `s.fA.fB...` — a scalar result is snapshotted into a fresh
+variable (value semantics, like `pto.local_array_get`); an aggregate result
+(nested array / struct) is the member-access lvalue itself.
+
+**Arguments:**
+
+| Name | Type | Description |
+|------|------|-------------|
+| `s` | `!pto.struct<...>` | The struct |
+| `path` | `DenseI64ArrayAttr` | Constant field indices, e.g. `[0, 2]` |
+
+**Results:** `value` — the field type at the end of `path`.
+
+**Constraints & Verification:**
+
+- `path` is non-empty; every index is in range at its level.
+- An intermediate index may only descend into a nested struct.
+- Result type must equal the field type at the end of `path`.
+
+**Basic Example:**
+
+```mlir
+%r = pto.struct_get %s[0]    : !pto.struct<f16, i8> -> f16      // r = s.f0;
+%v = pto.struct_get %s[0, 1]                                    // v = s.f0.f1;
+   : !pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>> -> i8
+```
+
+---
+
+##### `pto.struct_set` - Write a Field by Constant Path
+
+**Summary:** Write a value into the field reached by a constant `path` of field
+indices, descending through nested structs.
+
+**Semantics:** `s.fA.fB... = value;`
+
+**Arguments:**
+
+| Name | Type | Description |
+|------|------|-------------|
+| `s` | `!pto.struct<...>` | The struct |
+| `path` | `DenseI64ArrayAttr` | Constant field indices, e.g. `[0, 2]` |
+| `value` | `AnyType` | Value to write (must match the field type) |
+
+**Results:** None.
+
+**Constraints & Verification:**
+
+- `path` is non-empty; every index is in range at its level.
+- Value type must equal the field type at the end of `path`.
+- The terminal field must be copy-assignable: a `!pto.local_array` terminal is
+  rejected (a C array cannot be assigned). Use `pto.struct_get` to obtain the
+  array, then write elements with `pto.local_array_set`.
+
+**Basic Example:**
+
+```mlir
+pto.struct_set %s[1], %v : !pto.struct<f16, i8>, i8            // s.f1 = v;
+pto.struct_set %s[0, 0], %v                                    // s.f0.f0 = v;
+   : !pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>>, f32
 ```
 
 ---
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -340,10 +340,25 @@ func.func @make() -> !pto.struct<f32, i8> {
 } else { ... }
 ```
 
+**A function may not return a `!pto.struct` either**, even one that just passes
+an argument back:
+
+```mlir
+// Rejected: laundering. %r is no longer a pto.declare_struct result, so the
+// per-op check above cannot see that `return %r` publishes the caller's local.
+func.func private @passthrough(%s : !pto.struct<f32, i8>) -> !pto.struct<f32, i8> {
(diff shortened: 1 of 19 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #753 feat(pto): add !pto.struct heterogeneous aggregate type and ops
- Author: `zhangstevenunity`
- Base/Head: `main` / `claude/naughty-shirley-adce24`
- Head SHA: `20cb64c7373d`
- Trigger: 检测到新的 open PR
- Generated At: `2026-06-04T15:49:09Z`
- Status: completed

### Summary

发现 2 个 P2 级类型映射问题：新引入的 `!pto.struct` 在 EmitC 后端的命名和字段类型映射都不是一一对应，既会生成同名重复定义，也会把部分合法字段类型降成错误的 C++ 表示。

### Findings

1. **P2** 不同的 \`!pto.struct\` 类型会被降成同一个 C++ 结构体名 `lib/PTO/Transforms/PTOToEmitC.cpp`:`609`

`getStructTypeName()` 直接复用 `getEmitCScalarTypeToken()` 做名字编码，但这个 token 不是按 MLIR 类型一一映射的：例如 `i32` 和 `si32` 都会变成 `int32_t`，`i1` 还会落入默认分支变成 `float`。`collectStructTypes()` 是按 `StructType` 去重而不是按最终 C++ 名称去重，所以同一模块里一旦同时出现 `!pto.struct<i32>`/`!pto.struct<si32>`，或 `!pto.struct<i1>`/`!pto.struct<f32>`，这里会生成两份同名的 `struct PtoStruct_*` 定义，最终 C++ 会直接因为重定义而编译失败。

2. **P2** \`!pto.struct\` 接受了 EmitC 后端无法正确表示的标量字段类型 `lib/PTO/IR/PTO.cpp`:`10717`

`StructType::verify()` 现在只检查 `isIntOrFloat()`，因此 `!pto.struct<i1>`、`!pto.struct<i24>` 之类字段类型都会通过校验。但 EmitC 侧的字段渲染/命名只精确支持 `f16/bf16/f32/f64` 和 8/16/32/64 位整数；像 `i1`、`i24` 会在 `renderStructFieldDecl()` / `getEmitCScalarTypeToken()` 中退化成 `float`，而类型转换又把值当成 `i1`/`int32_t` 使用。这样 `pto.struct_get` / `pto.struct_set` 对这些“合法”字段会生成依赖隐式 float/int 转换的 C++，语义已经不再等价。这里需要把 verifier 收紧到后端实际支持的标量集合，或补齐这些类型的稳定映射。

zhangstevenunity:
## Rebased onto current main + fixed the red CI

Pushed 2 fix commits on top of the original, rebased from the old base (636 commits behind) onto current `main`.

### All three red CI jobs were one bug

`parseStructPath` declared `int64_t idx;` uninitialized. GCC cannot see through `parser.parseInteger`'s out-parameter, so `-Werror=maybe-uninitialized` rejected it. That single warning broke *Build ptoas-bin (x86_64)*, *(aarch64)* and vpto-sim-validation's "Build PTOAS" step. `build-and-test` stayed green only because it does not build with GCC + `-Werror`.

### Codex's two P2 findings — both real, both fixed

1. **Duplicate struct names.** `getStructTypeName` mangled from `getEmitCScalarTypeToken`, which is many-to-one (`i32` and `si32` both give `int32_t`; `i1` fell through to the `float` default). Two distinct `!pto.struct` types could therefore emit two `struct PtoStruct_*` definitions under one name, and the generated C++ would fail with a redefinition. Names are now mangled from the **MLIR type spelling** (`!pto.struct<f16, i8>` → `PtoStruct_f16_i8`), which is injective by construction. New test: `struct_name_collision_emitc.pto`.
2. **Verifier too loose.** `StructType::verify` only checked `isIntOrFloat()`, so `i1`/`i24` and the packed f8/f4 formats verified but were emitted as `float`, silently changing the field's width. Tightened to integer widths 8/16/32/64 and f16/bf16/f32/f64. New test: `struct_unrepresentable_scalar_invalid.pto`.

### Gemini's two HIGH findings are false positives

Both claim `walkStructPath` / `buildStructMemberChain` cannot compile because callers pass `getS().getType()` where `StructType` is expected. `StructType:$s` in ODS generates a `TypedValue<StructType>` accessor, so `.getType()` already returns `StructType` — no conversion is involved. The failing build log is the direct disproof: it reached `PTO.cpp:14393`, thousands of lines past `walkStructPath`. No change made.

### The rebase exposed a blocker — please read

On today's `main` this feature **assert-crashes on every `pto.struct` input**: `cast<TypedValue<emitc::LValueType>>` on an `OpResult`. Current MLIR models C++ lvalues explicitly — `emitc.variable` yields `!emitc.lvalue<T>`, and `emitc.member` both takes and returns an lvalue. The original lowering predates that model. Merging without this rebase would have shipped a compiler that crashes on the feature it adds.

The lowering is ported: the type converter maps `!pto.struct` to `!emitc.lvalue<!emitc.opaque<...>>`, `declare_struct` is the `emitc.variable` itself with no load, the member chain stays in lvalue form, and `struct_get`/`struct_set` are `emitc.load`/`emitc.assign`.

### Two constructs had to be dropped

Both are **unimplementable** against current EmitC, so the verifier now rejects them up front instead of letting the backend fail later:

- **`!pto.local_array` as a field type.** A field is reached with `emitc.member`, whose result must be an `!emitc.lvalue`, and `!emitc.lvalue` cannot wrap the `!emitc.array` an array field lowers to (`mlir-opt` rejects it outright). `emitc.subscript` also refuses an lvalue operand, so there is no way to spell the access. Pointer-decay would work only for rank 1 and fights the type converter.
- **A `get`/`set` path ending on a nested `!pto.struct`.** Returning a whole aggregate as an SSA value would copy it out of the parent struct. Nested fields are reached with a longer path (`%s[0, 0]`) instead, which lowers to a chained `s.f0.f0` — covered by the updated `struct_nested_emitc.pto`.

Nested `!pto.struct` fields themselves still work. Docs and ODS descriptions are updated to match; happy to revisit if you want array fields back via a different representation.

### Verification

Built locally with LLVM 21 + clang and `PTOAS_ENABLE_WERROR=ON`. Full lit suite run against a baseline build of `upstream/main`: **identical results — 128 pre-existing failures on both, zero regressions, zero deltas**. All 8 struct tests pass.

include/PTO/IR/PTOTypeDefs.td
```diff
@@ -303,6 +303,30 @@ def LocalArrayType : TypeDef<PTO_Dialect, "LocalArray"> {
   }];
 }
 
+// ---- !pto.struct<T0, T1, ..., Tn-1> ----
+// Scalar-unit heterogeneous aggregate. The heterogeneous dual of
+// !pto.local_array: it lowers to a plain C++ `struct { ... };` whose address is
+// decided by the host C++ compiler, never by PTO memory planning or
+// pto.pointer_cast. Fields are positionally numbered (f0, f1, ...), matching an
+// LLVM literal struct. A field type may be a scalar (int / float), a nested
+// !pto.local_array, or a nested !pto.struct — this is how nested / complex
```
zhangstevenunity: This comment contradicts `StructType::verify` added in this same PR. `isStructStorable` accepts only a scalar or a nested `!pto.struct`, so a `!pto.local_array` field is rejected with `is not scalar-storable` -- which is exactly what `test/lit/pto/struct_set_array_invalid.pto` asserts.

Please drop `a nested !pto.local_array` from this list. Line 321 has the same problem: the `// form: !pto.struct<f16, i8, !pto.local_array<4xf32>>` example shows a type the verifier refuses, so the one illustrative spelling in the ODS is an invalid one.

This is the ODS half of the desync; `docs/PTO_IR_manual.md` has it too (see the other inline comments).

docs/PTO_IR_manual.md
```diff
@@ -10405,6 +10469,102 @@ pto.local_array_set %m[%i, %j], %v
...
+
+---
+
+##### `pto.struct_get` - Read a Field by Constant Path
+
+**Summary:** Read the field reached by a constant `path` of field indices,
+descending through nested structs.
+
+**Semantics:** `s.fA.fB...` — a scalar result is snapshotted into a fresh
+variable (value semantics, like `pto.local_array_get`); an aggregate result
```
zhangstevenunity: Stale semantics -- there is no case where `struct_get` yields "the member-access lvalue itself".

Section 2.7 of this same file (added in this PR) states the path "must **end at a scalar**", and `verifyStructLeafIsScalar` enforces it: an aggregate result fails with `struct path must end at a scalar field`, covered by `test/lit/pto/struct_get_aggregate_invalid.pto`.

Suggest reducing this to the scalar case only: the result is snapshotted into a fresh variable, full stop.

docs/PTO_IR_manual.md
```diff
@@ -10405,6 +10469,102 @@ pto.local_array_set %m[%i, %j], %v
...
+- `path` is non-empty; every index is in range at its level.
+- An intermediate index may only descend into a nested struct.
+- Result type must equal the field type at the end of `path`.
+
+**Basic Example:**
+
+```mlir
+%r = pto.struct_get %s[0]    : !pto.struct<f16, i8> -> f16      // r = s.f0;
+%v = pto.struct_get %s[0, 1]                                    // v = s.f0.f1;
+   : !pto.struct<!pto.struct<f32, i8>, !pto.local_array<4xf32>> -> i8
```
zhangstevenunity: This example does not verify.

`!pto.local_array<4xf32>` as a struct field is rejected by `StructType::verify` with `is not scalar-storable`, so a reader who copies this hits an error on the *type* before `struct_get` is ever reached.

Suggest switching to a type the verifier accepts and that the tests actually exercise, e.g. `!pto.struct<!pto.struct<f32, i8>, i16>` with path `[0, 1]` returning `i8` -- the shape `test/lit/pto/struct_nested_emitc.pto` covers.

docs/PTO_IR_manual.md
```diff
@@ -10405,6 +10469,102 @@ pto.local_array_set %m[%i, %j], %v
...
+| `path` | `DenseI64ArrayAttr` | Constant field indices, e.g. `[0, 2]` |
+| `value` | `AnyType` | Value to write (must match the field type) |
+
+**Results:** None.
+
+**Constraints & Verification:**
+
+- `path` is non-empty; every index is in range at its level.
+- Value type must equal the field type at the end of `path`.
+- The terminal field must be copy-assignable: a `!pto.local_array` terminal is
```
zhangstevenunity: This entire constraint describes the design that was dropped.

There is no `!pto.local_array` terminal to reject, because an array cannot be a field at all. And the suggested workaround -- "use `pto.struct_get` to obtain the array, then write elements with `pto.local_array_set`" -- is not performable: `struct_get` must end at a scalar and cannot hand back an aggregate.

Replace with the rule the code actually enforces: the path must end at a scalar, and the value type must equal that field's type.

The example just below (line 10563) uses the same rejected `!pto.local_array` field type and needs the same fix as the `struct_get` example.

lib/PTO/Transforms/PTOToEmitC.cpp
```diff
@@ -14405,6 +14605,30 @@ struct EmitPTOManualPass
...
+        // content-derived !emitc.opaque tokens.
+        {
+          llvm::SetVector<pto::StructType> structDefs;
+          mop.walk([&](Operation *op) {
+            for (Type t : op->getResultTypes())
+              collectStructTypes(t, structDefs);
+            for (Value v : op->getOperands())
+              collectStructTypes(v.getType(), structDefs);
+            if (auto func = dyn_cast<func::FuncOp>(op)) {
+              for (Type t : func.getArgumentTypes())
```
zhangstevenunity, line 14670: Minor / P3, not blocking.

These two branches do fire for a struct-typed function signature, but such a module can never finish lowering, so the definition collected here is never emitted. I built this head and tried both directions:

- struct as a func **argument**: `error: failed to legalize operation 'func.func' that was explicitly marked illegal`
- struct as a func **result**: `error: 'emitc.return' op operand #0 must be type supported by EmitC, but got '!emitc.lvalue<!emitc.opaque<"PtoStruct_f16_i8">>'`

Both are clean rejections -- no crash, no miscompile -- which is why this is only P3. But the messages leak EmitC internals and never state the actual rule: a `!pto.struct` converts to an lvalue, so it cannot cross a function boundary. Worth either an explicit up-front diagnostic or, at minimum, a line in section 2.7 saying struct values are function-local. Right now nothing in the docs mentions the restriction, so the first person to try it gets `failed to legalize operation 'func.func'`.

test/lit/pto/struct_set_array_invalid.pto
```diff
@@ -0,0 +1,23 @@
...
+// RUN: not ptoas %s 2>&1 | FileCheck %s
+
+// A !pto.local_array cannot be a struct field. A field is reached with
+// emitc.member, whose result must be an !emitc.lvalue, and !emitc.lvalue cannot
+// wrap the !emitc.array an array field lowers to — so the access is not
+// expressible and the type verifier rejects it up front.
+
+module {
+  func.func @struct_array_field() {
+    %s = pto.declare_struct -> !pto.struct<!pto.local_array<4xf32>>
```
zhangstevenunity, line 18: Naming and coverage nit.

This file is named `struct_set_array_invalid` but it no longer exercises `pto.struct_set` at all -- the type verifier rejects the struct type on this line, so the test never gets as far as constructing a `struct_set`. Renaming to something like `struct_array_field_invalid.pto` would match what it checks.

Two small gaps worth closing while here:

1. Nothing covers the `descends into non-struct field` diagnostic in `walkStructPath` -- e.g. path `[0, 0]` on `!pto.struct<f32, i8>`. That is the one verifier branch with no test.
2. There is no parse/print round-trip test for the new type. All three positive tests go straight to EmitC text, so `StructType::parse` / `print` are only covered incidentally.

zhangstevenunity (commented):
Re-reviewed at head `1eeb62502` (rebased onto LLVM21 main, CI green). I built this exact head and ran `ptoas` on it rather than reading the diff alone. **The compiler side holds up -- I found no correctness bug.** What is left is a docs/ODS layer that still describes the *previous* design, including two manual examples that this PR's own verifier rejects.

### Verified by running the PR-head build

- Scalar and nested codegen match the lit expectations, and `struct_get` value semantics hold: `half v4 = v3.f0; v3.f0 = v4;`.
- Mangling from the MLIR type spelling keeps `i32` / `si32` / `ui32` structs distinctly named.
- Every scalar type the tightened verifier admits has an exact `getEmitCScalarTypeToken` spelling (f16/bf16/f32/f64, i8/16/32/64 signed+unsigned), so the trailing `return "float"` fallback is unreachable for struct fields. No silent narrowing.
- Struct-definition emission runs at line 14614, before `applyPartialConversion` at 14839, so the walk sees un-converted `pto.struct` types; nested-first ordering yields valid C++.
- `getEmitCVariableResultType` already passes an lvalue through unchanged, so `declare_struct` does not double-wrap.
- Empty path `[]` produces an actionable verifier error.

I specifically hunted for a reference-semantics miscompile, since `!pto.struct` converts to an `!emitc.lvalue`: I fed it an `scf.for` whose body yields a *different* struct than its init value. That is cleanly rejected (`failed to reconcile SCF structural types`), and the pass-through case lowers correctly. No silent aliasing bug on that path.

### P2 -- docs and the ODS comment still describe the abandoned `local_array`-field design

Two examples in `PTO_IR_manual.md` are IR that this PR's verifier refuses. A reader copying them gets `field 1 ... is not scalar-storable` -- precisely what `test/lit/pto/struct_set_array_invalid.pto` asserts. The `PTOTypeDefs.td` header comment still advertises `!pto.local_array` as a legal field type, contradicting `StructType::verify` added a few hundred lines away in the same PR. Per `.claude/rules/cross-layer-sync.md` these layers have to move together. Details inline. **The PR description carries the same stale claim** ("a nested local_array field can be indexed via local_array_get/set") and should be updated too.

### P3 -- non-blocking

Struct-typed function signatures are rejected with diagnostics that leak EmitC internals, and there are two small test gaps. Inline.

*(Procedural note: GitHub does not allow REQUEST_CHANGES on one's own PR, so this is filed as COMMENT -- but I would treat the doc corrections as merge-blocking, since they ship examples that do not verify.)*

zhangstevenunity:
## Fixed: struct function arguments (P1)

Confirmed and fixed in `947c9ca4`. Reproducer:

```mlir
func.func private @helper(%s : !pto.struct<f32, i8>, %v : f32) {
  pto.struct_set %s[0], %v : !pto.struct<f32, i8>, f32
  return
}
```
→ `error: failed to legalize operation 'func.func' that was explicitly marked illegal`

The diagnosis is exactly right. I verified both halves against LLVM 21 directly:

- `emitc.func` with an lvalue argument → `error: 'emitc.func' op cannot have lvalue type as argument`.
- Keeping `func.func` does not help either: it passes the MLIR verifier but `mlir-translate --mlir-to-cpp` then fails with `'func.func' op cannot emit lvalue type as argument type`.

So the lvalue conversion I introduced made struct parameters unrepresentable end to end.

### Fix: carry the struct as a pointer

The type converter now maps `!pto.struct` → `!emitc.ptr<!emitc.opaque<"PtoStruct_X">>`. A pointer is legal in a signature and still names the caller's storage.

Worth calling out: **by-value would have been a silent miscompile, not just an alternative.** `emitc.member` needs an lvalue, so a by-value parameter would have made the callee's `pto.struct_set` write into a copy and the caller would never see the update. The pointer is what makes the mutation semantics correct, not merely what makes it compile.

Locals keep the nicer form — `declare_struct` declares the storage and hands out its address, and the member chain unwraps that address-of and walks the variable directly, so a struct that never leaves the function still prints as `s.f0`. Only a genuine pointer root (a function argument) goes through `->`; `emitc.member_of_ptr` needs an lvalue *holding* the pointer rather than the raw pointer, so the argument is parked in a variable first.

Emitted C++ for the reproducer:

```c++
static AICORE void helper(PtoStruct_f32_i8* v1, float v2) {
  PtoStruct_f32_i8* v3;
  v3 = v1;
  v3->f0 = v2;
  return;
}

AICORE void caller(float v1) {
  PtoStruct_f32_i8 v2;
  PtoStruct_f32_i8* v3 = &v2;
  v2.f0 = v1;
  helper(v3, v1);
  return;
}
```

### Verification

New test `struct_func_arg_emitc.pto` covers both sides of the call. Full lit suite re-run against a baseline build of `upstream/main`: **128 failures on both, zero regressions, zero deltas**; all 9 struct tests pass.

One cosmetic leftover: a struct that is never passed anywhere still emits an unused `PtoStruct_X* vN = &vM;`. It is dead and any optimizer drops it, but say the word if you'd rather I suppress it.

zhangstevenunity:
## Fixed: stack storage escaping as a dangling pointer (P1)

Confirmed and fixed in `303f8a8f`. This one was worse than the report suggests — the return case did not fail, it **silently produced undefined behaviour**:

```mlir
func.func private @make(%v : f32) -> !pto.struct<f32, i8> {
  %s = pto.declare_struct -> !pto.struct<f32, i8>
  pto.struct_set %s[0], %v : !pto.struct<f32, i8>, f32
  return %s : !pto.struct<f32, i8>
}
```

`ptoas` exited **0** and emitted:

```c++
static AICORE PtoStruct_f32_i8* make(float v1) {
  PtoStruct_f32_i8 v2;
  v2.f0 = v1;
  return &v2;          // address of a dead frame
}
```

That compiles clean and is UB at run time. The `scf.if` yield case did fail, but only downstream and as an unhelpful `failed to legalize operation 'pto.struct_get'`, which points at the wrong op entirely.

Both are a direct consequence of carrying the struct as a pointer to stack storage (the fix for the previous P1), so the pointer model needs this guard to be sound.

### Fix

`pto.declare_struct` now rejects its result reaching a terminator. One rule covers every escape path — `func.return`, `scf.yield`, `cf.br` — and it reports at the declaration rather than downstream:

```
error: 'pto.declare_struct' op stack-local struct must not escape the scope that
declares it, but its value is passed to 'func.return', which would expose the
address of storage that is about to die; declare the struct in the outer scope
and pass it down as a function argument instead (pto.struct_set mutates in
place, so a struct never needs to be returned or yielded)
```

**This costs no expressiveness.** `pto.struct_set` mutates in place, so returning or yielding a struct is never necessary — declare it in the outer scope and pass it down; the callee writes through the pointer and the caller sees the result.

### Deliberately still allowed

The check is scoped to `pto.declare_struct` results, so returning a struct that arrived as a **function argument** stays legal — that is the caller's storage and it outlives the callee. Verified it still lowers correctly:

```c++
static AICORE PtoStruct_f32_i8* passthrough(PtoStruct_f32_i8* v1, float v2) {
  v3 = v1;
  v3->f0 = v2;
  return v1;           // caller's storage, not a local
}
```

Two exotic cases are rejected slightly conservatively: yielding a struct that was declared *outside* the region, and `cf.br` block arguments. Both are safe in principle but pointless given in-place mutation, and the diagnostic tells you to use the value directly. Happy to narrow it if you'd rather.

### Verification

Tests for both escape paths (`struct_escape_return_invalid.pto`, `struct_escape_yield_invalid.pto`). Full lit suite against a baseline build of `upstream/main`: **128 failures on both, zero regressions**; all 11 struct tests pass. Docs and ODS updated with the rule and the reasoning.

zhangstevenunity:
## Fixed: indirect escape past the verifier (P1)

Confirmed and fixed in `24c0fbea`. You're right, and the hole was **the carve-out I explicitly defended in my previous comment** — I need to retract that.

Last round I wrote that returning a struct which arrived as a function argument "stays legal — that is the caller's storage and it outlives the callee." That is true of the callee in isolation, and it is exactly the laundering vector. It hands the *caller* a struct value with no recoverable provenance:

```mlir
func.func private @passthrough(%s : !pto.struct<f32, i8>) -> !pto.struct<f32, i8> {
  return %s : !pto.struct<f32, i8>          // "safe" in isolation
}

func.func @leak(%v : f32) -> !pto.struct<f32, i8> {
  %s = pto.declare_struct -> !pto.struct<f32, i8>
  %r = func.call @passthrough(%s) : (...) -> !pto.struct<f32, i8>
  return %r : !pto.struct<f32, i8>          // %r is not a declare result -> unchecked
}
```

`ptoas` exited **0** and emitted:

```c++
AICORE PtoStruct_f32_i8* leak(float v1) {
  PtoStruct_f32_i8 v2;
  v2.f0 = v1;
  return passthrough(&v2);   // address of a dead frame
}
```

Same silent-UB class as the direct case, just routed around the check.

### Fix

A function result is the *only* place struct provenance can be laundered, so `validateStructNeverReturned` rejects `!pto.struct` as a function result type module-wide, run alongside the existing entry-function validation in `PTOToEmitC`.

Combined with the per-op check, that leaves exactly two ways to obtain a struct value — **declare it**, or **receive it as an argument**. Provenance is therefore always local and always visible to the verifier, which is what makes the direct-user check sound rather than merely usually-right.

The argument is inductive: every struct value now roots in either a live local (a declare, which cannot escape its scope) or a caller's storage (an argument, whose lifetime the caller guarantees). There is no third source, so there is no laundering path left.

Passing structs **down** into helpers is unaffected, and nothing is lost — `pto.struct_set` mutates in place, so a result is never needed.

### Verification

Escape matrix, all four now rejected, previously two of them silently emitted UB:

| case | before | after |
|---|---|---|
| return a local struct | exit 0, `return &v2;` | rejected |
| `scf.yield` out of `scf.if` | obscure legalize failure | rejected at the declare |
| **indirect via pass-through helper** | **exit 0, `return passthrough(&v2);`** | **rejected** |
| helper returning its own argument | allowed (the vector) | rejected |

New test `struct_escape_indirect_invalid.pto`. Full lit suite vs a baseline build of `upstream/main`: **128 failures on both, zero regressions**; 12/12 struct tests pass. Docs and ODS updated, and the diagnostic is plain ASCII so it renders correctly on non-UTF-8 consoles.

lib/PTO/IR/PTO.cpp
```diff
@@ -2799,6 +2890,58 @@ LogicalResult mlir::pto::validatePTOEntryFunctions(ModuleOp module) {
   return success();
 }
 
+// A !pto.struct is represented as a pointer to stack storage. Its provenance
+// therefore has to remain explicit: the value either comes directly from
+// pto.declare_struct or is received as a function entry argument. Operations
+// such as arith.select and scf.if must not manufacture another struct-typed SSA
+// result, because that alias hides the declaration from DeclareStructOp's
+// direct-use escape check. CFG block arguments may relay an existing value:
+// forwarding a declaration directly is already rejected because the branch is
```
zhangstevenunity, line 2899: Non-blocking -- the invariant is stated a touch too narrowly. A declare result can reach a struct block arg *without* going through a terminator: an `scf.while` init operand forwards its value into the before-region block arg, and that operand sits on the `scf.while` op itself (not a terminator), so `DeclareStructOp::verify()`'s terminator check does not fire on it.

The conclusion still holds, but for a different reason: to stay well-typed the loop must `scf.yield` a struct back in the after-region, and every producer of that yielded struct is already closed (a declare there reaches the yield terminator; any other op-result is now banned; only a function argument survives, whose storage outlives the loop). So the surviving block-arg cases are exactly the function-arg ones.

Might be worth rewording to something like "a declare can only reach a block arg along a path that also trips the terminator or op-result gate," which covers `scf.while` too.

zhangstevenunity (commented):
## Re-review at head f465cd4d ("close derived !pto.struct escape paths")

Reviewed the five commits added since my last pass at `1eeb6250`, focusing on the new commit `f465cd4d`, which generalizes the old `validateStructNeverReturned` (function-result ban only) into `validateStructProvenance`.

**Verdict: sound hardening, no correctness regression -- I would treat it as mergeable pending the in-flight GCC builds.** (Self-review, so this is a COMMENT rather than an approve.)

### What the new commit does, and why it is correct
The prior escape scheme had two gates: (1) `DeclareStructOp::verify()` rejects a declare result that reaches a terminator, and (2) a module-wide ban on `!pto.struct` function results. A real hole slipped between them: an `arith.select` / `scf.if` / call / cast whose *result* is `!pto.struct` is neither a declare result (so gate 1 misses the yield of the alias) nor a function result (so gate 2 misses it), so a `scf.yield %alias` could publish a pointer to branch-local storage at exit code 0. The new op-result gate closes exactly that, and the new negative test `struct_provenance_result_invalid.pto` reproduces it faithfully -- the `arith.select` is precisely what launders the declare past gate 1's terminator check, which is what makes it a genuine test of the new gate rather than the old one.

### Soundness of the resulting scheme
After this commit a `!pto.struct` SSA value can be produced only by (a) `pto.declare_struct` or (b) a block argument. Op results everywhere else are banned; function results are banned; declare results cannot reach a terminator. Block arguments are the only uninspected producer, and they are safe: a struct block arg can only carry function-argument provenance -- a declare forwarded through a `cf.br` hits gate 1 on the branch terminator, and `scf.for` struct iter-args force a struct *result* that the new gate rejects, so loop-carry is closed too.

### Prior finding resolved
My previous review's only finding -- the docs/ODS desync still advertising the abandoned `local_array`-as-field design -- is fully fixed: `PTOTypeDefs.td` now documents scalar-or-nested-struct fields and explicitly rejects `local_array`, the manual examples and section numbering (2.6/2.7/2.8) are consistent, and the op-count 110 -> 113 is correct.

### One non-blocking note
See the inline comment: the block-argument safety justification in the code comment is stated a touch too narrowly (an `scf.while` init operand is a non-terminator path that *can* forward a declare into a struct block arg). The conclusion still holds -- it is safe for a different reason -- but the stated invariant does not literally cover that path.

### CI
The definitive GCC `-Werror` wheel builds (`Build ptoas-bin x86_64/aarch64`) are still running. They were green at the prior head, and the delta here is a trivially safe rename plus a simple `module.walk`; I found no new uninitialized/unused/sign surface in the added C++. `vpto-sim-validation` is skipped (no sim-relevant change).

reedhecre:
## A5 板测成功
- 触发方式：`merged`
- 源码提交：`77f2fce1444e`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260722_130206_merged_pr753.log`
- 结果 TSV：`/root/ptoas-board-monitor-a5/logs/20260722_130206_merged_pr753.tsv`

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`77f2fce1444e`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260721_220111_merged_pr753.log`

### 失败用例
- `comm_p2p` (run, exit=1)

reedhecre:
## A3 板测失败详情：PR #753

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260721_220111_merged_pr753/npu_validation/CommSync/comm_p2p_binding_variants/main.cpp:122)
EZ9999[PID: 3167014] 2026-07-21-22:29:03.287.368 (EZ9999):  The error from device(chipId:0, dieId:0), serial number is 1021, there is an exception of aivec error, core id is 46, error code = 0x10, dump info: pc start: 0x124000000000, current: 0x124000000164, vec error info: 0x9e000000ab, mte error info: 0x6d0600009a, ifu error info: 0x23ba43c403cc0, ccu error info: 0x400001800000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600009a, fixp_error1 info: 0x6d, fsmId:1, tslot:7, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=0, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_binding_variants_kernel, fault kernel info ext=comm_p2p_binding_variants_kernel, program id=0, hash=16541545288638093324.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-21 22:29:04] ERROR: testcase failed (exit 1): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260721_220111_merged_pr753/npu_validation/CommSync/comm_p2p/main.cpp:122)
EZ9999[PID: 3187046] 2026-07-21-22:29:23.790.256 (EZ9999):  The error from device(chipId:0, dieId:0), serial number is 1022, there is an exception of aivec error, core id is 1, error code = 0x10, dump info: pc start: 0x124000000000, current: 0x124000000168, vec error info: 0xfa000000c1, mte error info: 0x9d06000011, ifu error info: 0x2298178444cc0, ccu error info: 0x40e008000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x6000011, fixp_error1 info: 0x9d, fsmId:0, tslot:7, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=0, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_kernel, fault kernel info ext=comm_p2p_kernel, program id=0, hash=6260346515940177617.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-21 22:29:25] ERROR: testcase failed (exit 1): comm_p2p
```

</details>

PR #835, opened 2026-06-18
feat: first stage of vmi

mouliangyu:


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #835 feat: first stage of vmi
- Author: `mouliangyu`
- Base/Head: `main` / `feature-vmi`
- Head SHA: `cc17114e5286`
- Trigger: PR 有新提交
- Generated At: `2026-07-14T16:34:03Z`
- Previous Head SHA: `a167c637f923`
- Status: failed at `codex-review` (exit=1)

### Summary

Review failed at stage `codex-review`: exit=1

### Findings

### Log Tail

```text
Updating files:  89% (3928/4413)
Updating files:  90% (3972/4413)
Updating files:  91% (4016/4413)
Updating files:  92% (4060/4413)
Updating files:  93% (4105/4413)
Updating files:  94% (4149/4413)
Updating files:  95% (4193/4413)
Updating files:  96% (4237/4413)
Updating files:  97% (4281/4413)
Updating files:  97% (4296/4413)
Updating files:  98% (4325/4413)
Updating files:  99% (4369/4413)
Updating files: 100% (4413/4413)
Updating files: 100% (4413/4413), done.
From https://github.com/hw-native-sys/PTOAS
 * [new ref]           refs/pull/835/head -> pr-835
From https://github.com/hw-native-sys/PTOAS
 * branch              main       -> FETCH_HEAD
Switched to branch 'pr-835'
cc17114e5286089024303adf7e10ea932fdabf07
fatal: origin/main...HEAD: no merge base
===== END STAGE clone rc=0 @ 2026-07-15 00:33:11 =====

===== STAGE codex-review @ 2026-07-15 00:33:11 =====
set -euo pipefail
cd '/tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/repo'
'codex' exec -C '/tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/repo' -s read-only -c 'model_provider="codereview"' -c 'model="gpt-5.4"' -c 'model_reasoning_effort="xhigh"' --output-schema '/tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/review_schema.json' -o '/tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/codex_last_message.json' --color never - < '/tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/review_prompt.txt'
[monitor] stage timeout: 1800s
OpenAI Codex v0.115.0 (research preview)
--------
workdir: /tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/repo
model: gpt-5.4
provider: codereview
approval: never
sandbox: read-only
reasoning effort: xhigh
reasoning summaries: none
session id: 019f617a-0a8d-7720-9d4b-d0d623590289
--------
user
你现在在审查 GitHub PR。

仓库：hw-native-sys/PTOAS
PR：#835 feat: first stage of vmi
作者：mouliangyu
base branch：origin/main
head branch：HEAD（当前已 checkout 到 PR head）

要求：
1. 只审查这个 PR 相对 origin/main 的改动，必要时可以看上下文文件。
2. 重点找真实的 correctness / regression / contract mismatch / CI / runtime / compatibility 问题。
3. 不要提纯风格建议，不要提低价值猜测。
4. 严格按优先级输出：
   - P1：高概率会导致错误结果、编译/运行失败、严重回归、发布阻断
   - P2：重要缺陷、行为回归、遗漏校验/测试、较大兼容性问题
   - P3：次要但明确可改的问题
5. 如果没有问题，summary 直接写：未检查到 PR #835 存在问题，并返回 findings=[]。
6. 如果有问题，summary 简洁概括，findings 里每条都要给出：
   - severity
   - title
   - body（说明为什么是问题，尽量具体）
   - file（尽量给相对路径）
   - line（能确定就填整数，否则 null）

建议先查看：
- git status --short
- git diff --stat origin/main...HEAD
- git diff --unified=80 origin/main...HEAD

最终输出必须严格匹配 JSON schema。

mcp startup: no servers
Reconnecting... 1/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1ee656d1415ce-SJC, request id: 8753371f-77dc-44ea-9435-f240c3de2239)
Reconnecting... 2/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1ee6c2f34707f-SJC, request id: 5ec60f7e-0659-4385-a8af-d04d30229461)
Reconnecting... 3/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1ee783d7cb8e7-SJC, request id: 80e3e91c-8b01-40d0-b40d-e1b2fb1c54b1)
Reconnecting... 4/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1ee873b16939b-SJC, request id: 02d1bf34-06df-4212-9944-8aef6acfc5f9)
Reconnecting... 5/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1ee96e8c8cfe1-SJC, request id: 48ebf7a6-14c9-4650-80b7-3b97eabc6261)
ERROR: unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b1eeba6832d4f9-SJC, request id: 8edd70ad-86fe-4459-bf0b-31c84c1a1e60
Warning: no last agent message; wrote empty content to /tmp/ptoas-pr-review-monitor/runs/20260715_003148_pr835/codex_last_message.json
===== END STAGE codex-review rc=1 @ 2026-07-15 00:34:03 =====
```

zhangstevenunity (commented):
Reviewed the core of this first VMI stage — the dialect (`VMI.cpp`, `VMITypeDefs`/`VMIOps.td`), `VMILayoutAssignment`, and `VMIToVPTO` — against the VPTO contracts.

The design holds up well: the closed-set layout model (`contiguous` / `deinterleaved={2,4}`) with a precise lane-map, the union-find layout assignment that hard-errors on genuine conflicts, and the two-tier "permissive verifier + strict lowering-capability gate" all read as deliberate, and unsupported cases are generally bailed explicitly rather than silently miscompiled.

Three concrete correctness issues below, all on currently-untested paths:
1. `addWhileConstraints` mis-aligns the two `scf.while` carry groups.
2. `truncf` hardcodes `rnd="R"`, emitting an illegal `pto.vcvt` for `f32 -> HiFloat8`.
3. Float `iota` builds a `FloatAttr` with mismatched (IEEEdouble) semantics.

(Layout-coalescing cost in loops, integer width conversion, and the histogram/half layout appear to be deliberately out of first-stage scope per the design docs, so not raised here.)

lib/PTO/Transforms/VMILayoutAssignment.cpp
```diff
@@ -0,0 +1,1330 @@
...
+    auto conditionOp =
+        dyn_cast<scf::ConditionOp>(whileOp.getBefore().front().getTerminator());
+    auto yieldOp = dyn_cast<scf::YieldOp>(afterBlock.getTerminator());
+
+    for (auto [index, init] : llvm::enumerate(inits)) {
+      Value anchor = init;
+      if (index < beforeArgs.size() &&
+          failed(uniteEquivalentValues(anchor, beforeArgs[index], whileOp)))
+        return failure();
+      if (conditionOp && index < conditionOp.getArgs().size() &&
```
zhangstevenunity: `addWhileConstraints` mis-aligns the two independent `scf.while` carry groups.

`scf.while` has two *separate* carries:
- before group: `inits[i]` == `beforeArgs[i]` == the after-region `scf.yield` operand `i`
- after group: `scf.condition` operand `j` == `afterArg[j]` == `result[j]`

These lists can differ in arity and element type (the before region computes the forwarded `scf.condition` operands). This loop anchors all five lists on the single `inits`-indexed `anchor`, so `conditionOp.getArgs()[index]` (L864), `afterBlock.getArgument(index)` (L868) and `whileOp.getResult(index)` (L876) get united against the wrong carry group. For any non-pass-through `scf.while` this yields a spurious "conflicting layouts" error, or silently forces one carry's layout onto an unrelated value. Only the degenerate 1-element pass-through is tested (`vmi_layout_assignment_scf_while.pto`).

Suggest two independent unions: the before group over `{inits[i], beforeArgs[i], yield.getOperand(i)}`, and the after group over `{conditionOp.getArgs()[j], afterArgs[j], results[j]}` — the after group must not be anchored on `inits`.

lib/PTO/Transforms/VMIToVPTO.cpp
```diff
@@ -0,0 +1,6269 @@
...
+
+    FailureOr<Value> sourceMask =
+        createAllTrueMaskForVReg(op.getLoc(), sourceType0, rewriter);
+    FailureOr<Value> resultMask =
+        createAllTrueMaskForVReg(op.getLoc(), resultType, rewriter);
+    if (failed(sourceMask) || failed(resultMask))
+      return rewriter.notifyMatchFailure(op,
+                                         "failed to build truncf masks");
+
+    StringAttr rnd = rewriter.getStringAttr("R");
```
zhangstevenunity, line 11129: `truncf` hardcodes `rnd = "R"`, but the 8-bit (P0-P3) branch is shared by fp8 *and* HiFloat8, and the `pto.vcvt` contract for `f32 -> HiFloat8` requires `rnd in {A, H}` (`lookupVcvtContract(F32, HiF8)` -> `allowedRnd="AH"`, enforced by `VcvtOp::verify`).

`VMITruncFOp::verify` accepts an `hif8` result and `checkSupportedTruncFShape` checks only result bit-width/factor (not element type), so `pto.vmi.truncf : ...f32 -> ...hif8` (deinterleaved=4 source) passes every VMI gate and then emits a `pto.vcvt {rnd="R"}` that fails its own verifier. No test covers `f32 -> hif8` (existing truncf tests are f32->f16 / f32->fp8E*).

Suggest selecting `rnd` from the target type (R for f8E4M3/f8E5M2/f16/bf16; A or H for HiFloat8), or rejecting HiFloat8 in `checkSupportedTruncFShape` with a clear unsupported error.

lib/PTO/Transforms/VMIToVPTO.cpp
```diff
@@ -0,0 +1,6269 @@
...
+                                            PatternRewriter &rewriter) {
+  if (auto intType = dyn_cast<IntegerType>(type)) {
+    return rewriter
+        .create<arith::ConstantOp>(loc, IntegerAttr::get(intType, value))
+        .getResult();
+  }
+  if (auto floatType = dyn_cast<FloatType>(type)) {
+    return rewriter
+        .create<arith::ConstantOp>(
+            loc, FloatAttr::get(floatType,
```
zhangstevenunity: `llvm::APFloat(static_cast<double>(value))` carries IEEEdouble semantics, but `floatType` here is f16/bf16/f32. `FloatAttr::get(Type, APFloat)` requires the value's semantics to match the type (`verifyInvariants` checks `&type.getFloatSemantics() == &value.getSemantics()`), so this asserts in an asserts-enabled build (and yields an invalid attribute otherwise). Reachable for any non-f64 float `iota` with a nonzero lane offset (e.g. `vmi_to_vpto_iota_f16_deint2_asc`) and via the `factor`/`zero` constants in `createIotaDeinterleavedChunk`.

Use the semantics-aware overload `FloatAttr::get(floatType, static_cast<double>(value))` (which converts to the type's semantics), or build `APFloat(floatType.getFloatSemantics(), ...)` and round.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
```diff
@@ -0,0 +1,1650 @@
...
+        return builder.create<VMIMulIOp>(loc, ty, lhs, rhs).getResult();
+      };
+      (void)lowerVecScalar(vop, builder, createLegacy);
+      continue;
+    }
+
+    if (auto vop = dyn_cast<VMIMaxSOp>(op)) {
+      auto createLegacy = [&](Location loc, Type ty, Value lhs,
+                              Value rhs) -> Value {
+        return builder.create<VMIMaxFOp>(loc, ty, lhs, rhs).getResult();
```
zhangstevenunity: **P2 (verifier/lowering contract mismatch).** `VMIMaxSOp`/`VMIMinSOp` unconditionally build `VMIMaxFOp`/`VMIMinFOp`, but the shared vector-scalar verifier `verifyVMIVectorScalarOp` (VMI.cpp) accepts integer element types too, and there is no `VMIMaxIOp`/`VMIMinIOp`. Compare `vadds`/`vmuls` just above, which branch on `isFloatType`.

So `%r = pto.vmi.vmaxs %v, %s : !pto.vmi.vreg<128xi32>, i32` passes the op verifier, then hard-fails in `VMIMaxFOp::verify` with a misleading "requires f16, bf16, or f32". Either add an integer max/min legacy op and branch like `vadds`/`vmuls`, or reject integer `vmaxs`/`vmins` in the op verifier so the diagnostic points at the real cause.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
```diff
@@ -0,0 +1,1650 @@
...
+/// Unified mask and pmode are intentionally discarded.
+///
+/// \p createLegacy is a callable `(Location, Type, Value, Value) -> Value`
+/// that emits the legacy binary op.
+template <typename UnifiedOp>
+static LogicalResult
+lowerBinaryIgnoringMask(
+    UnifiedOp op,
+    function_ref<Value(Location, Type, Value, Value)> createLegacy) {
+  if (hasMergePmode(op))
```
zhangstevenunity, line 268: **P2 (silent miscompile when reachable).** `lowerBinaryIgnoringMask`/`lowerMaskedUnary` bail only on `pmode == "merge"`. For `pmode == "zero"` -- and for an absent pmode, since `getPmodeOrDefault` returns `"zero"` -- the mask is dropped and *all* lanes are computed. But the op contract (VMIOps.td: `"zero": inactive lanes are zeroed`) and `verifyVMIVariadicPmodeMask` both allow mask + `pmode=zero`, so a masked `vadd/.../vneg {pmode="zero"}` yields non-zero inactive lanes instead of 0, with no diagnostic.

Secondary: `getPmodeOrDefault` defaults to `"zero"` while VMIOps.td documents the default as `"merge"` -- an absent-pmode masked op is thus treated as zero here *and* has its mask dropped. If these masked/pmode forms are not yet meant to be emitted, please gate them in the verifier/validator rather than silently discarding the mask.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
```diff
@@ -0,0 +1,1650 @@
...
+      (void)lowerVecScalar(vop, builder, createLegacy);
+      continue;
+    }
+
+    if (auto vop = dyn_cast<VMIShrSOp>(op)) {
+      Type elemType = getVMIElementType(vop.getSrc());
+      auto createLegacy = [&](Location loc, Type ty, Value lhs,
+                              Value rhs) -> Value {
+        auto intType = cast<IntegerType>(elemType);
+        if (intType.isUnsigned())
```
zhangstevenunity: **P2/P3 (sign-semantics inconsistency -> silent wrong result).** Signless right-shift is lowered as arithmetic (sign-filling): `intType.isUnsigned()` is false for a signless element, so this builds `VMIShRSIOp`. The VMI sign convention (VMI.cpp:104-134) states "signless integers are treated as equivalent to UNSIGNED only" precisely to avoid "silent sign-extension bugs when a producer happens to emit a signless value" -- which is exactly what happens for a signless `vshr`/`vshrs` with the high bit set. Passes.td documents "shrsi otherwise", so the lowering and the convention disagree; please reconcile (route through the sign-semantics helper, treating signless as logical / `shrui`). Same at the `vshrs` site (~1562).

lib/PTO/Transforms/VMILayoutRematerialize.cpp
```diff
@@ -0,0 +1,392 @@
...
+          continue;
+        }
+
+        if (auto ensure = dyn_cast<VMIEnsureMaskLayoutOp>(op)) {
+          changed |= tryReplaceMaskEnsure(ensure);
+          continue;
+        }
+
+        if (auto ensure = dyn_cast<VMIEnsureMaskGranularityOp>(op))
+          changed |= tryReplaceMaskEnsure(ensure);
```
zhangstevenunity, line 383: **P2 (use-after-free / UB).** Missing `continue` after this `VMIEnsureMaskGranularityOp` branch. `tryReplaceMaskEnsure` calls `ensure.erase()` on success (~line 343); then line 381 does `dyn_cast<VMITruncIOp>(op)` on the freed `Operation*` (reading its TypeID). Every other branch here has a `continue` for exactly this reason. The two op types are mutually exclusive, so it is functionally benign in release builds, but it is a real use-after-free that will trip ASan/sanitizer CI. Add `continue;` after line 379.

tools/ptoas/ObjectEmission.cpp
```diff
@@ -442,17 +470,39 @@ static bool compileDeviceLLVMToObject(llvm::StringRef llPath,
...
+    break;
+  case BishengVFAutoSyncMode::Global:
+    args.push_back("-mllvm");
+    args.push_back("-cce-vf-auto-sync=global");
+    break;
+  }
+  // Enabling vector MI scheduling deliberately omits this argument instead of
+  // passing `=1`, so Bisheng retains the default behavior of the selected
+  // toolchain version.
+  if (!enableBishengVecMISched) {
```
zhangstevenunity, line 496: **P2 (default device-codegen change for ALL VPTO output, not just VMI).** The previously-unconditional `-mllvm -cce-vf-auto-sync=global` is now emitted only when `--bisheng-vf-auto-sync` is passed (default `Unspecified` omits it -> toolchain default), and the new default additionally disables the vector MI scheduler (`--cce-aicore-vec-misched=0`) and always adds `--cce-long-scbz=true`. `compileDeviceLLVMToObject` is the common device-object path for every VPTO fatobj (A3/A5, VMI or not), so this changes emitted objects for existing kernels. These are sync/scheduling-relevant flags, and `remote-npu-validation` is skipped on this PR, so the change is only host-simulator-validated (which typically does not model VF sync hazards). Please confirm flipping these defaults for the non-VMI VPTO path is intended and validated on-device.

zhangstevenunity (commented):
**Review: "first stage of VMI"**

I reviewed the behavior-changing surface of this PR -- CLI/pipeline, packaging/wrapper, device object emission, the PTODSL frontend, and the VMI layout + lowering passes -- with several parallel deep-dives. The ~1080 test files and the 129-op dialect surface I sampled rather than read line-by-line.

**Overall:** a large, genuinely well-documented first stage (14 design docs + an 11-file VMI ISA spec), backed by a real numeric-validation harness (~95 `vmi_new` golden/compare cases) and a green `vpto-sim-validation` run. I did **not** find a P1 in the reviewed slices, and several high-risk areas checked out sound: the i16 loop-counter narrowing bound analysis; VMIToVPTO fails *closed* (partial conversion marks unhandled VMI ops illegal + a no-residual verifier) rather than silently miscompiling; the wrapper's path resolution across build/install/wheel trees; and the DSL IfExp/branch-literal/liveness changes.

**Main themes to weigh (details in inline comments):**

1. **No kill-switch.** Commit `075905927` removed the `--enable-vmi=false` escape hatch, so the whole VMI pipeline -- plus a new `createInlinerPass()`, LICM, loop-counter narrowing, and vcvt normalization -- is now unconditionally on for every `--pto-backend=vpto` compile. For a first-stage 96k-line middle-end, losing the fallback means any regression is unmitigatable without a rebuild and can't be A/B-bisected. I'd keep the disable flag through stabilization.

2. **Default device-codegen change for ALL VPTO output** (inline on `ObjectEmission.cpp`): VF auto-sync no longer forced to `global`, vector MI scheduler now disabled by default, `--cce-long-scbz=true` always added. `remote-npu-validation` is skipped, so this is host-simulator-only -- which typically does not model VF sync hazards.

3. **Correctness gaps in the always-on lowering** (inline on `VMILowerUnifiedToLegacy.cpp`): integer `vmaxs`/`vmins` lowered to a float op (hard verifier error); masked binary/unary ops drop the mask under `pmode="zero"` (silent wrong lanes); signless right-shift sign-fills against the codebase's own `signless == unsigned` convention. The file header still claims it runs "under the opt-in --vmi-two-stage-lowering flag" -- stale; it's always-on, which is what turns these shortcuts from opt-in into default behavior.

4. **A use-after-free** in `VMILayoutRematerialize` (inline): benign in release builds but will trip sanitizer/ASan CI.

5. **DSL frontend** (inline): duplicate `vintlv`/`vdintlv` defs make the PR's new implementations dead code; the static subscript-index resolver is narrower than the bound resolver and turns some previously-tracing runtime loops into hard errors.

**Lower priority:** dead helpers `rewriteDataTypes`/`insertDataUseMaterializations` in `VMILayoutAssignment` (never called); a missing `scf.while` after-region-only layout/granularity constraint in `VMILayoutAssignment`/`VMIMaskGranularityAssignment` (fail-loud via the scf verifier, not silent); `VPTONormalizeEquivalentVcvt` assumes even/odd-lane equivalence for the `US_*`/`E2B_*` load-dist modes (I couldn't confirm those modes actually duplicate lane pairs -- worth double-checking each); and a stray repro committed under `lib/` (inline).

**Process nit:** the PR body is empty for a 1,223-file / 181-commit change. A short description pointing at `docs/designs/vmi-introduction.md` and the validation status would help reviewers a lot.

None of this needs to block continued iteration, but I'd restore the `--enable-vmi` off-switch and fix the integer `vmaxs`/`vmins` + the use-after-free before this lands widely.

ptodsl/ptodsl/_ops.py
```diff
@@ -1519,6 +1523,32 @@ def pdintlv_b32(lhs, rhs):
...
+    op = op_ctor(
+        lhs_type,
+        lhs_type,
+        unwrap_surface_value(lhs),
+        unwrap_surface_value(rhs),
+    )
+    return wrap_surface_value(op.low), wrap_surface_value(op.high)
+
+
+def vintlv(lhs, rhs):
```
zhangstevenunity: **P2 (latent -- this new code is dead).** These new `vintlv`/`vdintlv` (backed by `_vreg_pair_op`, adding a `lhs.type == rhs.type` check + `_infer_vreg_metadata`) are shadowed by identically-named defs later in this file (~2112/2124). In Python the later def wins, so `pto.vintlv`/`vdintlv` keep their old behavior and this code never runs (it is also imported twice in pto.py). No immediate miscompile -- but if someone deletes the old pair to "fix the duplicate", this untested path activates, and it unpacks results via `op.low`/`op.high` vs the live version's `.results`. Please delete one pair and keep the intended one.

ptodsl/ptodsl/_ast_rewrite.py
```diff
@@ -219,6 +302,250 @@ def _name_info(node) -> _NameInfo:
...
+    index_values = _static_index_values(node.slice, static_iters)
+    if index_values is None:
+        return set()
+    return {
+        _SubscriptSlot(node.value.id, index)
+        for index in index_values
+    }
+
+
+def _static_index_values(node, static_iters):
```
zhangstevenunity, line 482: **P2 (new hard-errors on previously-tracing code; `ast_rewrite` is on by default).** `_static_index_values` treats a subscript index as static only for an int `Constant`, a `USub` of one, or a bare `static_range` IV name. It does not handle `BinOp`, module globals, closures, or `const_expr` params -- yet the sibling bound resolver `_eval_static_int` handles `BinOp` + `static_env`. Since `_rewrite_for` raises on the first `invalid_stores` entry for every runtime `range()` loop, patterns that previously traced now raise `PTODSLAstRewriteError` with a misleading message. Examples inside `for c in range(cols):` -> `buf[NHEADS] = v` (global index), `accs[h+1] = ...` under a `static_range(4)` (fully static!), or a slot carry sized by a `const_expr`. Suggest routing subscript indices through the same static evaluator used for bounds, or at least fixing the diagnostic to name the real cause. (Gap is untested -- all new tests use literal indices.)

lib/TileOps/issue_518_repro.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) 2026 Huawei Technologies Co., Ltd.
```
zhangstevenunity: **P3 (repo hygiene / testing policy).** This looks like a one-off repro/benchmark committed under `lib/TileOps/`. It is not referenced by any CMake/test and violates the project testing-and-examples rule ("Avoid: temporary scripts committed outside test/ or test/samples"). Please move it into `test/samples/` (or convert to a proper regression test) or drop it.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
mouliangyu: Fixed in `28c6523b7`. Added integer legacy `VMIMinIOp` and `VMIMaxIOp`, including verifiers, layout propagation and rematerialization support, and VMIToVPTO lowering. `vmaxs` and `vmins` now dispatch on the element type like `vadds` and `vmuls`; integer forms lower end-to-end to `pto.vmax` and `pto.vmin`. Added `vmi_lower_vmaxs_vmins_integer.pto` for both stages.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
mouliangyu: This behavior is intentional for the two-stage lowering contract. The bottom compute op owns zero-mode semantics, so zero-mode lowering does not synthesize a select or merge chain here. `pmode=merge` remains unsupported and returns failure, while an absent pmode follows zero mode in this lowering. We are therefore keeping the lowering behavior. The surface wording around the default mode should be kept aligned with this contract separately.

lib/PTO/Transforms/VMILowerUnifiedToLegacy.cpp
mouliangyu: Fixed in `28c6523b7`. Both `vshr` and `vshrs` now select `shrsi` only for an explicitly signed integer type; signless and unsigned types select `shrui`, matching the VMI sign convention. The regression test now covers signed, unsigned, and signless vector and vector-scalar cases.

lib/PTO/Transforms/VMILayoutRematerialize.cpp
mouliangyu: Fixed in `28c6523b7`. The `VMIEnsureMaskGranularityOp` branch now immediately continues after `tryReplaceMaskEnsure`, so an erased operation is never passed to the following `dyn_cast<VMITruncIOp>`.

tools/ptoas/ObjectEmission.cpp
mouliangyu: Confirmed intentional for the common VPTO device-object path, not an accidental VMI-only change. The intended defaults are to stop forcing global VF auto-sync unless explicitly requested, disable the vector MI scheduler, and enable long SCBZ for all VPTO output. Regressions in existing non-VMI VPTO kernels are in scope under these defaults and should be addressed directly rather than hidden behind VMI scoping or a fallback path.

ptodsl/ptodsl/_ops.py
mouliangyu: Fixed in `28c6523b7`. Removed the shadowed first `vintlv` and `vdintlv` pair together with the now-unused `_vreg_pair_op` helper. Only the previously live definitions remain, so there is no latent behavior change from deleting the wrong duplicate later.

ptodsl/ptodsl/_ast_rewrite.py
mouliangyu: Agreed that `_static_index_values` is narrower than `_eval_static_int`. This is independent from the lowering and control-flow fixes in `28c6523b7` and remains a follow-up: the intended fix is to reuse the same static evaluator for subscript indices and add BinOp, global, closure, and const-expression regression cases. We should preserve the invalid-store check rather than weaken it only to make the gate pass.

lib/TileOps/issue_518_repro.py
mouliangyu: Fixed in `28c6523b7`. Removed `lib/TileOps/issue_518_repro.py`; it was an unreferenced one-off repro and no production or test target depended on it.

mouliangyu:
Review follow-up summary for commit `28c6523b7`:

- Fixed integer `vmaxs` and `vmins` with dedicated legacy `maxi` and `mini` ops and end-to-end VPTO tests.
- Fixed signless `vshr` and `vshrs` to use unsigned semantics.
- Fixed the rematerialization use-after-free.
- Moved `scf.for` and `scf.while` carry relationships into shared `VMIControlFlowSupport`. The two independent `scf.while` carry sequences are now modeled separately, with an asymmetric regression test.
- Removed the dead duplicate PTODSL interleave wrappers and the stray issue 518 repro.

The following are intentional design decisions:

- The VMI pipeline remains always-on for VPTO; we do not plan to restore a VMI kill-switch.
- The device codegen defaults apply to all VPTO output by design. Existing VPTO regressions under the new defaults are in scope and should be fixed directly.
- Zero-mode lowering relies on the bottom op zero-mode contract; merge mode remains unsupported in the two-stage lowering.

The cached first return operands in `VMILayoutAssignment` are intentional. A `func.return` has operands but no SSA result to serve as the equivalence anchor for a function result slot. The first return operand provides that anchor for additional return sites. Ordinary operations already have `OpResult` values and do not need this cache. When call results exist, function result types come from those call results; otherwise the assigned type of the cached return value is used.

The static subscript evaluator gap is acknowledged as a separate follow-up. The fix should reuse the existing static evaluator and add BinOp, global, closure, and const-expression coverage without weakening invalid-store detection.

Validation completed locally: `pto-test-opt` built successfully, 3 new or updated focused tests passed, and 6 related existing control-flow and min/max tests passed.

zhangstevenunity:
Thanks for the fast turnaround on `28c6523b7` -- I re-reviewed the fix commit and it addresses the lowering + control-flow findings cleanly. Verified:

- **Integer `vmaxs`/`vmins`**: correct end-to-end -- new `VMIMinIOp`/`VMIMaxIOp` with verifiers, `isFloatType` dispatch in the lowering, VMIToVPTO patterns to `Vmin`/`Vmax`, the supported-op gate, and rematerialize support, plus the two-stage regression test.
- **Signless `vshr`/`vshrs`**: `!isSigned()` now routes signless -> `shrui` (logical), matching the sign convention; good that the test covers signed/unsigned/signless.
- **`VMILayoutRematerialize` use-after-free**: `continue` added -- resolved.
- **`scf.while` after-region asymmetry**: the new `VMIControlFlowSupport` correctly models the before-carry and after-carry as independent sequences (and nicely consolidates both passes onto the shared helper); regression test added.
- **Duplicate `vintlv`/`vdintlv`** and the stray `issue_518_repro.py`: removed.

On the three you've dispositioned rather than changed -- all reasonable:

- **zero-pmode mask** (`VMILowerUnifiedToLegacy.cpp:268`): understood that the bottom compute op owns zero-mode semantics in the two-stage contract. The one thing worth pinning is a golden/numeric case with a masked `pmode="zero"` binary/unary op whose inactive lanes are asserted `== 0`, so that contract is guaranteed by a test rather than by convention; and +1 on aligning the `getPmodeOrDefault`=`"zero"` default with the VMIOps.td `"merge"` wording.
- **ObjectEmission defaults**: clear, thanks -- deliberate global VPTO change with regressions owned directly. Since `remote-npu-validation` is skipped on CI, an on-device spot-check of one existing non-VMI kernel before/after would be worth doing, as these are sync/scheduling flags the host sim doesn't model.
- **`_ast_rewrite` static index resolver**: agreed it's a separate follow-up; keeping the invalid-store check is the right call.

Nothing blocking from my side. (Minor leftover: `rewriteDataTypes` / `insertDataUseMaterializations` in `VMILayoutAssignment` are still defined-but-unused after the refactor -- safe to drop whenever you next touch that file.)

reedhecre:
## A5 板测失败
- 触发方式：`merged`
- 源码提交：`4870cd36ac03`
- 结果汇总：OK `0` / FAIL `0` / SKIP `0`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260724_100408_merged_pr835.log`
- 失败阶段：`build-ptoas` / exit=1

### 日志尾部
```text
_HIDDEN_FLAG - Success
-- Performing Test CXX_SUPPORTS_SUGGEST_OVERRIDE_FLAG
-- Performing Test CXX_SUPPORTS_SUGGEST_OVERRIDE_FLAG - Success
-- Performing Test CXX_WSUGGEST_OVERRIDE_ALLOWS_ONLY_FINAL
-- Performing Test CXX_WSUGGEST_OVERRIDE_ALLOWS_ONLY_FINAL - Success
-- Performing Test C_WCOMMENT_ALLOWS_LINE_WRAP
-- Performing Test C_WCOMMENT_ALLOWS_LINE_WRAP - Failed
-- Performing Test C_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG
-- Performing Test C_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG - Failed
-- Performing Test CXX_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG
-- Performing Test CXX_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG - Success
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS - Failed
-- Looking for os_signpost_interval_begin
-- Looking for os_signpost_interval_begin - not found
-- Found MLIR 
-- MLIR CMake Dir: /home/ptoas-board-monitor-a5/llvm-project-vpto-llvm21/build-shared/lib/cmake/mlir
-- Found Python3: /usr/bin/python3 (found version "3.11.6") found components: Interpreter Development.Module 
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Found pybind11: /usr/local/lib/python3.11/site-packages/pybind11/include (found version "2.13.6")
-- Performing Test LLVM_LINKER_SUPPORTS_B_SYMBOLIC_FUNCTIONS
-- Performing Test LLVM_LINKER_SUPPORTS_B_SYMBOLIC_FUNCTIONS - Success
-- Found Python3: /usr/bin/python3 (found version "3.11.6") found components: Interpreter 
-- Building pto lit tests
-- Configuring done (1.9s)
-- Generating done (0.0s)
-- Build files have been written to: /tmp/ptoas-board-monitor-a5/runs/20260724_100408_merged_pr835/repo/build
[0/2] Re-checking globbed directories...
ninja: error: unknown target 'ptoas', did you mean 'ptobc'?
===== END STAGE build-ptoas rc=1 @ 2026-07-24 10:04:17 =====
```

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`4870cd36ac03`
- 结果汇总：OK `0` / FAIL `0` / SKIP `0`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260723_190310_merged_pr835.log`
- 失败阶段：`build-ptoas` / exit=1

### 日志尾部
```text
NTATION_FLAG - Success
-- Performing Test CXX_SUPPORTS_MISLEADING_INDENTATION_FLAG
-- Performing Test CXX_SUPPORTS_MISLEADING_INDENTATION_FLAG - Success
-- Performing Test C_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG
-- Performing Test C_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG - Success
-- Performing Test CXX_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG
-- Performing Test CXX_SUPPORTS_CTAD_MAYBE_UNSPPORTED_FLAG - Success
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS - Failed
-- Looking for os_signpost_interval_begin
-- Looking for os_signpost_interval_begin - not found
-- Found MLIR 
-- MLIR CMake Dir: /home/zhongxuan/ptoas-board-monitor/cache/llvm-project-vpto-llvm21/build-shared/lib/cmake/mlir
-- Found Python3: /usr/bin/python3 (found version "3.10.9") found components: Interpreter Development.Module 
-- Performing Test HAS_FLTO_THIN
-- Performing Test HAS_FLTO_THIN - Failed
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Failed
-- Found pybind11: /home/zhongxuan/.local/lib/python3.10/site-packages/pybind11/include (found version "2.13.6")
-- Performing Test LLVM_LINKER_SUPPORTS_B_SYMBOLIC_FUNCTIONS
-- Performing Test LLVM_LINKER_SUPPORTS_B_SYMBOLIC_FUNCTIONS - Success
-- Found Python3: /usr/bin/python3 (found version "3.10.9") found components: Interpreter 
-- Building pto lit tests
-- Configuring done
-- Generating done
-- Build files have been written to: /home/zhongxuan/ptoas-board-monitor/runtime/runs/20260723_190310_merged_pr835/repo/build
ninja: Entering directory `/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260723_190310_merged_pr835/repo/build'
[0/2] Re-checking globbed directories...
ninja: error: unknown target 'ptoas', did you mean 'ptobc'?
===== END STAGE build-ptoas rc=1 @ 2026-07-23 19:04:43 =====
```

reedhecre:
## A3 板测失败
- 触发方式：`manual`
- 源码提交：`4870cd36ac03`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260723_192145_manual_pr835.log`
- 手动指令：`/run a3`
- 触发人：`zhongxuan`

### 失败用例
- `comm_p2p` (run, exit=1)

reedhecre:
## A3 板测失败详情：PR #835

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260723_192145_manual_pr835/npu_validation/CommSync/comm_p2p_binding_variants/main.cpp:122)
EZ9999[PID: 3069713] 2026-07-23-19:53:02.157.401 (EZ9999):  The error from device(chipId:1, dieId:1), serial number is 286, there is an exception of aivec error, core id is 17, error code = 0x10, dump info: pc start: 0x124600000000, current: 0x124600000164, vec error info: 0xe300000095, mte error info: 0x6f0600006f, ifu error info: 0x3015685020000, ccu error info: 0x40e008000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600006f, fixp_error1 info: 0x6f, fsmId:1, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=3, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_binding_variants_kernel, fault kernel info ext=comm_p2p_binding_variants_kernel, program id=0, hash=16541545288638093324.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-23 19:53:03] ERROR: testcase failed (exit 1): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260723_192145_manual_pr835/npu_validation/CommSync/comm_p2p/main.cpp:122)
EZ9999[PID: 3084438] 2026-07-23-19:53:21.927.069 (EZ9999):  The error from device(chipId:1, dieId:1), serial number is 287, there is an exception of aivec error, core id is 22, error code = 0x10, dump info: pc start: 0x124600000000, current: 0x124600000168, vec error info: 0xe300000095, mte error info: 0x6f0600006f, ifu error info: 0x3015685020000, ccu error info: 0x40a414030000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600006f, fixp_error1 info: 0x6f, fsmId:1, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=3, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_kernel, fault kernel info ext=comm_p2p_kernel, program id=0, hash=6260346515940177617.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-23 19:53:24] ERROR: testcase failed (exit 1): comm_p2p
```

</details>

reedhecre:
## A5 板测失败
- 触发方式：`manual`
- 源码提交：`4870cd36ac03`
- 结果汇总：OK `209` / FAIL `17` / SKIP `1`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260724_102201_manual_pr835.log`
- 手动指令：`/run a5`
- 触发人：`root`

### 失败用例
- `sels` (run, exit=139)
- `sel` (run, exit=2)
- `scatter` (run, exit=2)
- `rowexpandsub` (run, exit=2)
- `rowexpandmul` (run, exit=2)
- `rowexpanddiv` (run, exit=2)
- `rmsnorm` (run, exit=2)
- `post_rmsnorm` (run, exit=2)
- `partmin` (run, exit=2)
- `mscatter` (run, exit=2)
- `movfp_fixpipe_reuse` (run, exit=2)
- `mgather` (run, exit=2)
- `comm_p2p_binding_variants` (run, exit=2)
- `comm_p2p` (run, exit=2)
- `cmps` (run, exit=2)
- `cmp` (run, exit=2)

reedhecre:
## A5 板测失败详情：PR #835

<details><summary>sels</summary>

`stage=run` `info=exit=139`

```text
./test/npu_validation/scripts/run_remote_npu_validation.sh: line 848: 4151866 Segmentation fault      (core dumped) LD_LIBRARY_PATH="${LD_LIBRARY_PATH_NPU}" ./build/${testcase}
[2026-07-24 10:53:30] ERROR: testcase failed (exit 139): sels
```

</details>

<details><summary>sel</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v4.bin vs v4.bin, max diff=5.792495250701904 at idx=298 (golden=2.8790016174316406, out=-2.9134936332702637, dtype=float32)
[ERROR] compare failed
[2026-07-24 10:53:42] ERROR: testcase failed (exit 2): sel
```

</details>

<details><summary>scatter</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v3.bin vs v3.bin, max diff=3.6916218996047974 at idx=10 (golden=2.568453073501587, out=-1.1231688261032104, dtype=float32)
[ERROR] compare failed
[2026-07-24 10:53:52] ERROR: testcase failed (exit 2): scatter
```

</details>

<details><summary>rowexpandsub</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v3.bin vs v3.bin, max diff=5.565198540687561 at idx=202 (golden=1.555970311164856, out=-4.009228229522705, dtype=float32)
[ERROR] compare failed
[2026-07-24 10:55:28] ERROR: testcase failed (exit 2): rowexpandsub
```

</details>

<details><summary>rowexpandmul</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v3.bin vs v3.bin, max diff=14.702873229980469 at idx=206 (golden=8.388465881347656, out=-6.3144073486328125, dtype=float32)
[ERROR] compare failed
[2026-07-24 10:55:40] ERROR: testcase failed (exit 2): rowexpandmul
```

</details>

<details><summary>rowexpanddiv</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v3.bin vs v3.bin, max diff=18.47865390777588 at idx=963 (golden=10.328862190246582, out=-8.149791717529297, dtype=float32)
[ERROR] compare failed
[2026-07-24 10:56:51] ERROR: testcase failed (exit 2): rowexpanddiv
```

</details>

<details><summary>rope_kv_cache</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch (bf16 golden_v1.bin vs v1.bin): max ulp diff=30463 at idx=26720 (golden_bits=48012, out_bits=15218, golden=-0.0042724609375, out=0.003692626953125)
[ERROR] compare failed
[2026-07-24 10:59:17] ERROR: testcase failed (exit 2): rope_kv_cache
```

</details>

<details><summary>rmsnorm</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch (bf16 golden_v2.bin vs v2.bin): max ulp diff=4 at idx=73741 (golden_bits=15203, out_bits=15207, golden=0.0034637451171875, out=0.0035247802734375)
[ERROR] compare failed
[2026-07-24 10:59:28] ERROR: testcase failed (exit 2): rmsnorm
```

</details>

<details><summary>post_rmsnorm</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch (bf16 golden_v2.bin vs v2.bin): max ulp diff=3 at idx=8821 (golden_bits=15613, out_bits=15616, golden=0.0308837890625, out=0.03125)
[ERROR] compare failed
[2026-07-24 11:01:29] ERROR: testcase failed (exit 2): post_rmsnorm
```

</details>

<details><summary>partmin</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v3.bin vs v3.bin, max diff=nan at idx=112 (golden=-0.0, out=nan, dtype=float16)
[ERROR] compare failed
[2026-07-24 11:06:54] ERROR: testcase failed (exit 2): partmin
```

</details>

<details><summary>mscatter</summary>

`stage=run` `info=exit=2`

```text
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:104:25: error: cannot cast from type 'GlobalTensor<int32_t, pto::Shape<1, 1, 1, 32, 32>, pto::Stride<1024, 1024, 1024, 32, 1>, pto::Layout::ND>' (aka 'GlobalTensor<int, Shape<1, 1, 1, 32, 32>, Stride<1024, 1024, 1024, 32, 1>, pto::Layout::ND>') to pointer type '__gm__ int32_t *' (aka '__gm__ int *')
  __gm__ int32_t* v17 = (__gm__ int32_t*) v16;
                        ^~~~~~~~~~~~~~~~~~~~~
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:32:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/pto-inst.hpp:30:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:18:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr_impl.hpp:301:
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:461:24: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    using T = typename TileSrc::DType;
                       ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:2055:5: note: in instantiation of function template specialization 'pto::MSCATTER_IMPL<pto::Coalesce::Row, pto::ScatterAtomicOp::None, pto::ScatterOOB::Undefined, pto::ScatterConflict::Last, __gm__ int *, __ubuf__ int *, __ubuf__ int *>' requested here
    MSCATTER_IMPL<Mode>(dst, src, indexes);
    ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:137:3: note: in instantiation of function template specialization 'pto::MSCATTER<pto::Coalesce::Row, __gm__ int *, __ubuf__ int *, __ubuf__ int *>' requested here
  MSCATTER<pto::Coalesce::Row>(v17, v28, v29);
  ^
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:32:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/pto-inst.hpp:30:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:18:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr_impl.hpp:301:
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:462:27: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    using TIdx = typename TileIdx::DType;
                          ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:466:60: error: member reference base type '__gm__ int *' is not a structure or union
    __gm__ T *tablePtr = reinterpret_cast<__gm__ T *>(table.data());
                                                      ~~~~~^~~~~
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:468:35: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    constexpr int kSrcValidRowS = TileSrc::ValidRow;
                                  ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:469:35: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    constexpr int kSrcValidColS = TileSrc::ValidCol;
                                  ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:473:35: error: member reference base type '__ubuf__ int *' is not a structure or union
    const uint32_t validRows = src.GetValidRow();
                               ~~~^~~~~~~~~~~~
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:474:35: error: member reference base type '__ubuf__ int *' is not a structure or union
    const uint32_t validCols = src.GetValidCol();
                               ~~~^~~~~~~~~~~~
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:477:37: error: type '__gm__ int *' cannot be used prior to '::' because it has no members
        using TableShape = typename GlobalTable::Shape;
                                    ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:480:63: error: member reference base type '__gm__ int *' is not a structure or union
        const uint32_t tableRows = static_cast<uint32_t>(table.GetShape(GlobalTensorDim::DIM_3));
                                                         ~~~~~^~~~~~~~~
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:482:26: error: member reference base type '__ubuf__ int *' is not a structure or union
            tablePtr, src.data(), indices.data(), validRows, validCols, tableRows);
                      ~~~^~~~~
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:394:24: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    using T = typename TileSrc::DType;
                       ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:464:5: note: in instantiation of function template specialization 'pto::MScatterCheck<pto::Coalesce::Row, pto::ScatterAtomicOp::None, __gm__ int *, __ubuf__ int *, __ubuf__ int *>' requested here
    MScatterCheck<Mode, Atomic>(table, src, indices);
    ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:2055:5: note: in instantiation of function template specialization 'pto::MSCATTER_IMPL<pto::Coalesce::Row, pto::ScatterAtomicOp::None, pto::ScatterOOB::Undefined, pto::ScatterConflict::Last, __gm__ int *, __ubuf__ int *, __ubuf__ int *>' requested here
    MSCATTER_IMPL<Mode>(dst, src, indexes);
    ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:137:3: note: in instantiation of function template specialization 'pto::MSCATTER<pto::Coalesce::Row, __gm__ int *, __ubuf__ int *, __ubuf__ int *>' requested here
  MSCATTER<pto::Coalesce::Row>(v17, v28, v29);
  ^
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mscatter/mscatter/mscatter_kernel.cpp:32:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/pto-inst.hpp:30:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:18:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr_impl.hpp:301:
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:395:27: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    using TIdx = typename TileIdx::DType;
                          ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MScatter.hpp:404:43: error: type '__gm__ int *' cannot be used prior to '::' because it has no members
    static_assert(std::is_same_v<typename GlobalTable::DType, __gm__ T>,
                                          ^
14 errors generated.
gmake[2]: *** [CMakeFiles/mscatter_kernel.dir/build.make:76: CMakeFiles/mscatter_kernel.dir/mscatter_kernel.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/mscatter_kernel.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
[2026-07-24 11:09:52] ERROR: testcase failed (exit 2): mscatter
```

</details>

<details><summary>movfp_fixpipe_reuse</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Mismatch: golden_v5.bin vs v5.bin, max diff=nan at idx=8 (golden=-35.25, out=nan, dtype=float16)
[ERROR] Mismatch: golden_v6.bin vs v6.bin, max diff=nan at idx=120 (golden=-8.3125, out=nan, dtype=float16)
[ERROR] Mismatch: golden_v7.bin vs v7.bin, max diff=nan at idx=172 (golden=-17.5, out=nan, dtype=float16)
[ERROR] compare failed
[2026-07-24 11:10:44] ERROR: testcase failed (exit 2): movfp_fixpipe_reuse
```

</details>

<details><summary>mgather</summary>

`stage=run` `info=exit=2`

```text
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mgather/mgather/mgather_kernel.cpp:92:25: error: cannot cast from type 'GlobalTensor<int32_t, pto::Shape<1, 1, 1, 32, 32>, pto::Stride<1024, 1024, 1024, 32, 1>, pto::Layout::ND>' (aka 'GlobalTensor<int, Shape<1, 1, 1, 32, 32>, Stride<1024, 1024, 1024, 32, 1>, pto::Layout::ND>') to pointer type '__gm__ int32_t *' (aka '__gm__ int *')
  __gm__ int32_t* v11 = (__gm__ int32_t*) v10;
                        ^~~~~~~~~~~~~~~~~~~~~
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mgather/mgather/mgather_kernel.cpp:32:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/pto-inst.hpp:30:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:18:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr_impl.hpp:300:
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MGather.hpp:513:24: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    using T = typename TileDst::DType;
                       ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:2004:5: note: in instantiation of function template specialization 'pto::MGATHER_IMPL<pto::Coalesce::Row, pto::GatherOOB::Undefined, __ubuf__ int *, __gm__ int *, __ubuf__ int *>' requested here
    MGATHER_IMPL<CMode>(dst, src, indexes);
    ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mgather/mgather/mgather_kernel.cpp:130:3: note: in instantiation of function template specialization 'pto::MGATHER<pto::Coalesce::Row, __ubuf__ int *, __gm__ int *, __ubuf__ int *>' requested here
  MGATHER<pto::Coalesce::Row>(v26, v11, v25);
  ^
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/Mgather/mgather/mgather_kernel.cpp:32:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/pto-inst.hpp:30:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr.hpp:18:
In file included from /tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/common/pto_instr_impl.hpp:300:
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/payload/pto-isa/include/pto/npu/a5/MGather.hpp:515:19: error: type '__ubuf__ int *' cannot be used prior to '::' because it has no members
    if constexpr (TileDst::Loc == TileType::Mat) {
                  ^
3 errors generated.
gmake[2]: *** [CMakeFiles/mgather_kernel.dir/build.make:76: CMakeFiles/mgather_kernel.dir/mgather_kernel.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/mgather_kernel.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
[2026-07-24 11:11:08] ERROR: testcase failed (exit 2): mgather
```

</details>

`stage=run` `info=exit=2`

```text
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/CommSync/comm_p2p_binding_variants/comm_p2p_binding_variants_kernel.cpp:151:16: error: the range of 1st parameter must be [4, 6]
  pipe_barrier(PIPE_FIX);
               ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/CommSync/comm_p2p_binding_variants/comm_p2p_binding_variants_kernel.cpp:56:18: note: expanded from macro 'PIPE_FIX'
#define PIPE_FIX PIPE_M
                 ^
1 error generated.
gmake[2]: *** [CMakeFiles/comm_p2p_binding_variants_kernel.dir/build.make:76: CMakeFiles/comm_p2p_binding_variants_kernel.dir/comm_p2p_binding_variants_kernel.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/comm_p2p_binding_variants_kernel.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
[2026-07-24 11:22:16] ERROR: testcase failed (exit 2): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=2`

```text
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/CommSync/comm_p2p/comm_p2p_kernel.cpp:147:16: error: the range of 1st parameter must be [4, 6]
  pipe_barrier(PIPE_FIX);
               ^
/tmp/ptoas-board-monitor-a5/runs/20260724_102201_manual_pr835/npu_validation/CommSync/comm_p2p/comm_p2p_kernel.cpp:56:18: note: expanded from macro 'PIPE_FIX'
#define PIPE_FIX PIPE_M
                 ^
1 error generated.
gmake[2]: *** [CMakeFiles/comm_p2p_kernel.dir/build.make:76: CMakeFiles/comm_p2p_kernel.dir/comm_p2p_kernel.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/comm_p2p_kernel.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
[2026-07-24 11:22:18] ERROR: testcase failed (exit 2): comm_p2p
```

</details>

<details><summary>cmps</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Packed mask mismatch: golden_v2.bin vs v2.bin, idx=4 (golden=98, out=0)
[ERROR] compare failed
[2026-07-24 11:26:25] ERROR: testcase failed (exit 2): cmps
```

</details>

<details><summary>cmp</summary>

`stage=run` `info=exit=2`

```text
[ERROR] Packed mask mismatch: golden_v3.bin vs v3.bin, idx=4 (golden=49, out=0)
[ERROR] compare failed
[2026-07-24 11:26:35] ERROR: testcase failed (exit 2): cmp
```

</details>

Issue #861, opened 2026-06-24
MGATHER row mode races on a2a3: PTOAS inserts no MTE2->S for the scalar index read (missing SyncMacroModel)


Little-oil:
## Summary

On a2a3, `pto.mgather` in **row** mode produces wrong gathered rows or hangs the
AICore (507018), nondeterministically across runs (same case flips between hang
and wrong-values = a data race). **Elem** mode works. Root cause is a
cross-pipe synchronization that **PTOAS should insert but currently does not**.

## Root cause (PTOAS sync modeling)

MGATHER reads its **index tile on the SCALAR (S) pipe** (`idxPtr[r]`) to compute
GM addresses. But `MGatherOp::getPipe()` returns a single pipe = `PIPE_MTE2`
(`include/PTO/IR/PTOOps.td:2856`), and the sync translator
(`PTOIRTranslator::getOpPipeline`) models **all** of mgather's operand reads as
happening on that one pipe (MTE2).

The `idx` tile is produced by `pto.tload` (MTE2). So the sync solver sees
producer MTE2 -> consumer MTE2 and inserts only an **MTE2<->MTE2 `pipe_barrier`**
— it never inserts **MTE2 -> S**. The scalar idx read therefore races the idx
DMA -> stale/garbage indices -> wrong rows or out-of-bounds addresses (507018).

This is the same class of op as `TScatter`/`TGather` (which also read indices on
the scalar pipe) — and those **already have a `SyncMacroModel`**
(`lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp`). `MGatherOp` has none.

`MGatherElemImpl` "works" only because the pto-isa template currently hand-codes
`PtoSetWaitFlag<PIPE_MTE2, PIPE_S>()`; `MGatherRowImpl` does not. That manual
sync is the **wrong layer** — it should be PTOAS's job.

## Proposed fix (PTOAS owns the sync)

Add a `SyncMacroModel` for `MGatherOp`, mirroring `getTScatterSyncMacroModel` /
`getTGatherSyncMacroModel`, so the solver learns the real per-pipe usage and
inserts the MTE2->S wait automatically.

`lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp` (anonymous namespace):

```cpp
std::optional<SyncMacroModel> getMGatherSyncMacroModel(pto::MGatherOp op) {
  // A5 lowers via the vector path (getPipe()==PIPE_V); no scalar idx loop.
[...]

PR #865, opened 2026-06-24
Fix mgather sync insertion macro inference

FangRui0:
主要变更：
为 pto.mgather 建立更精确的 macro sync model，覆盖 A2/A3、A5、GM->UB、GM->L1、Row/Elem 等路径。
修复 A2/A3 MGATHER<Row> 实际走 PIPE_S + PIPE_MTE2，但 InsertSync 漏建模导致缺少 MTE2 -> S 同步的问题。
从 bind_tile / pointer_cast 恢复 lowered memref 上丢失的 valid_shape + layout 元数据，避免 Row/Elem coalesce 推断错误。
同步修复 EmitC 默认 coalesce 推断，确保生成的 MGATHER<Row/Elem> 与 InsertSync pipe model 一致。

### DIFF
--- a/lib/PTO/IR/PTO.cpp
+++ b/lib/PTO/IR/PTO.cpp
@@ -8499,12 +8499,14 @@ LogicalResult MGatherOp::verify() {
                                              "dst")))
     return failure();
 
-  std::optional<pto::Coalesce> explicitCoalesce;
-  if (auto coalesceAttr = getCoalesceAttr())
-    explicitCoalesce = coalesceAttr.getValue();
+  auto coalesceAttr = getCoalesceAttr();
+  if (!coalesceAttr)
+    return emitOpError(
+        "expects mgather to specify an explicit coalesce attribute (row or "
+        "elem)");
 
   if (failed(verifyMGatherMScatterTileShape(getOperation(), dstTy, idxTy, "dst",
-                                            explicitCoalesce)))
+                                            coalesceAttr.getValue())))
     return failure();
 
   return success();
--- a/lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
+++ b/lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
@@ -266,38 +266,6 @@ static std::optional<SmallVector<Value, 2>> lookupMGatherValidDims(Value value) 
   return std::nullopt;
 }
 
-static std::optional<pto::TileBufConfigAttr>
-lookupMGatherTileConfig(Value value) {
-  if (auto bind = value.getDefiningOp<pto::BindTileOp>())
-    return bind.getConfig();
-  if (auto pc = value.getDefiningOp<pto::PointerCastOp>()) {
-    if (auto config = pc.getConfig())
-      return *config;
-    return std::nullopt;
-  }
-  if (auto subview = value.getDefiningOp<memref::SubViewOp>())
-    return lookupMGatherTileConfig(subview.getSource());
-  if (auto cast = value.getDefiningOp<memref::ReinterpretCastOp>())
-    return lookupMGatherTileConfig(cast.getSource());
-  if (auto cast = value.getDefiningOp<memref::CastOp>())
-    return lookupMGatherTileConfig(cast.getSource());
-
-  if (auto regionResult = dyn_cast<OpResult>(value)) {
-    if (auto fusionRegion = dyn_cast<pto::FusionRegionOp>(regionResult.getOwner())) {
-      auto yieldOp =
-          dyn_cast<pto::YieldOp>(fusionRegion.getBody().front().getTerminator());
-      if (!yieldOp)
-        return std::nullopt;
-      unsigned resultIndex = regionResult.getResultNumber();
-      if (resultIndex >= yieldOp.getNumOperands())
-        return std::nullopt;
-      return lookupMGatherTileConfig(yieldOp.getOperand(resultIndex));
-    }
-  }
-
-  return std::nullopt;
-}
-
 static std::optional<SmallVector<int64_t>>
 getMGatherOperandShape(Value value, bool useValidShape = true) {
   auto shape = getMGatherOperandShapeFromType(value.getType(), useValidShape);
@@ -323,84 +291,6 @@ getMGatherOperandShape(Value value, bool useValidShape = true) {
   return shape;
 }
 
-struct MGatherShapeLayoutInfo {
-  SmallVector<int64_t, 2> shape;
-  bool rowMajor = false;
-  bool colMajor = false;
-};
-
-static std::optional<MGatherShapeLayoutInfo>
-getMGatherShapeLayoutInfo(Value value) {
-  auto shape = getMGatherOperandShape(value);
-  if (!shape || shape->size() != 2)
-    return std::nullopt;
-
-  MGatherShapeLayoutInfo info;
-  info.shape.assign(shape->begin(), shape->end());
-
-  if (auto tileTy = dyn_cast<pto::TileBufType>(value.getType())) {
-    int32_t blayout = tileTy.getBLayoutValueI32();
-    info.rowMajor = blayout == static_cast<int32_t>(pto::BLayout::RowMajor);
-    info.colMajor = blayout == static_cast<int32_t>(pto::BLayout::ColMajor);
-  }
-
-  if (auto config = lookupMGatherTileConfig(value)) {
-    auto blayoutAttr = dyn_cast<pto::BLayoutAttr>(config->getBLayout());
-    if (!blayoutAttr)
-      return info;
-    info.rowMajor = blayoutAttr.getValue() == pto::BLayout::RowMajor;
-    info.colMajor = blayoutAttr.getValue() == pto::BLayout::ColMajor;
-    return info;
-  }
-
-  if (auto memRefTy = dyn_cast<MemRefType>(value.getType())) {
-    SmallVector<int64_t, 4> strides;
-    int64_t offset = ShapedType::kDynamic;
-    if (failed(getStridesAndOffset(memRefTy, strides, offset)) ||
-        strides.size() != 2)
-      return std::nullopt;
-    info.rowMajor = strides[1] == 1;
-    info.colMajor = strides[0] == 1;
-  }
-
-  return info;
-}
-
-// Infer the effective Coalesce when the attribute is absent. Mirrors the
-// isRowCoalescedMGatherIndexType heuristic from PTOToEmitC: row gather has a
-// unit-extent index (valid [1,R] row-major or [R,1] col-major), otherwise Elem.
-// InsertSync runs after PTOViewToMemref / buffer materialization, where tile
-// valid_shape + layout metadata may live on pto.bind_tile/pointer_cast rather
-// than on the bare memref type. Recover that value-chain metadata first, and
-// only fall back to memref physical shape/strides when no PTO tile metadata is
-// available.
-static pto::Coalesce inferMGatherCoalesce(pto::MGatherOp op) {
-  if (auto coalesceAttr = op.getCoalesceAttr())
-    return coalesceAttr.getValue();
-
-  auto dataInfo = getMGatherShapeLayoutInfo(op.getDst());
-  auto idxInfo = getMGatherShapeLayoutInfo(op.getIdx());
-  if (!dataInfo || !idxInfo)
-    return pto::Coalesce::Elem;
-
-  auto isUnit = [](int64_t d) {
-    return d == 0 || d == 1;
-  };
-  auto compat = [](int64_t a, int64_t b) {
-    return a != ::mlir::ShapedType::kDynamic &&
-           b != ::mlir::ShapedType::kDynamic && a == b;
-  };
-
-  const bool rowCoalesce1xR =
-      idxInfo->rowMajor && isUnit(idxInfo->shape[0]) &&
-      compat(idxInfo->shape[1], dataInfo->shape[0]);
-  const bool rowCoalesceRx1 =
-      idxInfo->colMajor && compat(idxInfo->shape[0], dataInfo->shape[0]) &&
-      isUnit(idxInfo->shape[1]);
-  return (rowCoalesce1xR || rowCoalesceRx1) ? pto::Coalesce::Row
-                                            : pto::Coalesce::Elem;
-}
-
 // MGather is a macro-like library call whose internal pipe usage depends on the
 // destination address space, the coalesce mode, and the target arch. Model each
 // reachable lowering path as SyncMacroPhases so InsertSyncAnalysis can derive the
@@ -418,8 +308,12 @@ std::optional<SyncMacroModel> getMGatherSyncMacroModel(pto::MGatherOp op) {
   auto dstSpace = getMGatherOperandAddressSpace(op.getDst().getType());
   const bool isGm2L1 = dstSpace && *dstSpace == pto::AddressSpace::MAT;
 
+  auto coalesceAttr = op.getCoalesceAttr();
+  if (!coalesceAttr)
+    return std::nullopt;
+
   const PTOArch arch = getTargetArch(op.getOperation());
-  const pto::Coalesce coalesce = inferMGatherCoalesce(op);
+  const pto::Coalesce coalesce = coalesceAttr.getValue();
 
   SyncMacroModel model;
 
--- a/lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
+++ b/lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
@@ -8,6 +8,8 @@
 
 #include "PTO/Transforms/InsertSync/SyncMacroModel.h"
 #include "PTO/IR/PTO.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/Matchers.h"
 
 using namespace mlir;
 using namespace mlir::pto;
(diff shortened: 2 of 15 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #865 Fix mgather sync insertion macro inference
- Author: `FangRui0`
- Base/Head: `main` / `codex/mgather-sync-insertion-fix`
- Head SHA: `ab5ea4abc378`
- Trigger: PR 有新提交
- Generated At: `2026-06-30T04:59:01Z`
- Previous Head SHA: `d37fd4d33748`
- Status: completed

### Summary

PR #865 fixes InsertSync-side MGATHER modeling, but it also regresses implicit MSCATTER coalesce inference for dynamic shapes and leaves the GraphSyncSolver path on the old single-pipe MGATHER behavior.

### Findings

1. **P2** Implicit MSCATTER row inference regresses for dynamic valid shapes `lib/PTO/Transforms/PTOToEmitC.cpp`:`6326`

`PTOMScatterToMSCATTER` still infers `coalesce` when the attribute is omitted, but the new helper it calls now rejects dynamic extents (`?`) instead of treating them as compatible. That means a valid row-form scatter like idx valid shape `[1, %rows]` or `[%rows, 1]` with runtime `%rows` is still accepted by the verifier, yet lowering can no longer prove `Row` and falls through to `MSCATTER<pto::Coalesce::Elem>`. This changes the emitted intrinsic overload and can silently change results for existing dynamic-shape kernels that relied on inferred row coalescing.

2. **P2** MGATHER macro sync fix does not cover the GraphSyncSolver path `lib/PTO/Transforms/GraphSyncSolver/SyncSolverIRTranslator.cpp`:`186`

The new phase-aware MGATHER modeling is only wired into InsertSync. `--enable-graph-sync-solver` still translates every `OpPipeInterface` op as a single `pipeRead = pipeWrite = op.getPipe()`, so MGATHER's hidden S/V/MTE2 phases remain invisible there. As a result, the original missing waits are still present when GraphSyncSolver is enabled for row/elem GM->UB cases and the GM->L1 variants, even though this PR fixes the InsertSync path. The lack of any `_gss` companion coverage in the PR makes this supported mode easy to miss.

lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
```diff
@@ -193,6 +195,325 @@ std::optional<SyncMacroModel> getTGatherSyncMacroModel(pto::TGatherOp op) {
...
+  const PTOArch arch = getTargetArch(op.getOperation());
+  const pto::Coalesce coalesce = inferMGatherCoalesce(op);
+
+  SyncMacroModel model;
+
+  if (isGm2L1) {
+    // P1/P2: GM -> L1 (Row + Elem). S reads GM idx; MTE2 DMAs GM -> L1 into dst.
+    // Elem additionally writes the GM scratch (owned by S in the template) before
+    // the MTE2 bulk copy reads it, so model scratch as a S def / MTE2 use.
+    addPhase(model, PipelineType::PIPE_S, ValueRange{},
```
zhangstevenunity: [Blocking] This S phase passes empty `defValues`, so `scratch` is modeled as an MTE2 *read* only (the `mte2Uses` a few lines down). But `MGatherOp::getEffects()` (PTO.cpp:12954-12959) deliberately marks `scratch` as a *Write* ("the op clobbers scratch ... model it as a write"), and the macro path bypasses `getEffects` entirely: in PTOIRTranslator the `getSyncMacroModel()` branch (:411) precedes -- and so wins over -- the `OpPipeInterface` branch (:415), whose `UpdatePTOOpInfo()` is the only consumer of `getEffects` (:626-639). So the scratch write dependency is silently dropped, which also contradicts the comment two lines up ("model scratch as a S def / MTE2 use").

Consequence: two GM->L1 Elem gathers sharing one `scratch` (e.g. reused across a loop), or any op that reads `scratch` after the gather, are left un-synced -- both gathers look like scratch *reads* to InsertSync, so the WAR/WAW is invisible and iter N+1's internal S-write can race iter N's still-in-flight MTE2-read. This is a soundness regression vs the pre-PR single-pipe (getEffects-based) model.

Fix: in the Elem branch, give this S phase scratch as a def, e.g.
  `addPhase(model, PipelineType::PIPE_S, ValueRange{op.getScratch()}, ValueRange{op.getIdx()});`

lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
```diff
@@ -193,6 +195,325 @@ std::optional<SyncMacroModel> getTGatherSyncMacroModel(pto::TGatherOp op) {
...
+// destination address space, the coalesce mode, and the target arch. Model each
+// reachable lowering path as SyncMacroPhases so InsertSyncAnalysis can derive the
+// real cross-pipe sync (notably the MTE2->S wait for the scalar index read that
+// MGatherOp::getPipe() hides by reporting a single pipe).
+//
+// Also reserve pto-isa's fixed internal event ids through hiddenEvents so
+// compiler-generated sync around the macro does not reuse EVENT_ID0 on pipe
+// pairs already consumed inside the library implementation.
+//
+// See docs/designs/2026-06-24-mgather-sync-insertion-plan.md.
```
zhangstevenunity: This comment references `docs/designs/2026-06-24-mgather-sync-insertion-plan.md`, but that file is not added by this PR and does not exist in the repo. Please add the design doc or drop the reference.

lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
```diff
@@ -193,6 +195,325 @@ std::optional<SyncMacroModel> getTGatherSyncMacroModel(pto::TGatherOp op) {
...
+  return shape;
+}
+
+struct MGatherShapeLayoutInfo {
+  SmallVector<int64_t, 2> shape;
+  bool rowMajor = false;
+  bool colMajor = false;
+};
+
+static std::optional<MGatherShapeLayoutInfo>
```
zhangstevenunity: This helper -- together with `lookupMGatherValidDims` / `lookupMGatherTileConfig` / `inferMGatherCoalesce` -- is a near-verbatim copy of `getGatherScatterShapeLayoutInfo` & friends in PTOToEmitC.cpp. The two copies MUST stay semantically identical: InsertSync's pipe model (here) and the emitted `MGATHER<Row/Elem>` (PTOToEmitC) are both derived from this inference, so any drift means InsertSync syncs a different pipe than the code actually runs on. They already differ in the `TileBufType` branch (the EmitC version returns on the type's validShape/blayout; this one additionally applies bind_tile validDims + config). It is latent today because both reach the memref branch at their respective pipeline stages, but please factor out a single shared helper to prevent future divergence.

zhangstevenunity (changes requested):
Reviewed at head `91fcde7a` (deep correctness pass on the InsertSync macro model).

Mechanism is sound: modeling `pto.mgather` as macro phases so InsertSync can derive the MTE2->S idx-read wait that `MGatherOp::getPipe()` hides is the right fix for #861, and the model faithfully reproduces `MGatherOp::getEffects()` on every path EXCEPT GM->L1 Elem (see inline). One blocking regression; the rest are non-blocking.

Note: the Codex bot comment is stale -- it ran at head `9410655f` (2026-06-24); current head is `91fcde7a`. 2 of its 3 findings are already resolved here and I verified them:
- [FIXED] P1 (coalesce inference ignored valid_shape+layout): now recovered via `lookupMGatherValidDims` / `lookupMGatherTileConfig` / strides; `mgather_padded_valid_shape_coalesce_sync.pto` covers padded-physical 1xR->Row and Rx1-row-major->Elem.
- [FIXED] #3 (A5 1x1 unreachable, shape read only from TileBufType): `getMGatherOperandShapeFromType` now handles MemRefType; `mgather_gm2ub_a5_elem_1x1_memref_macro_sync_model.pto` exercises the memref form.
- [OPEN]  the scratch P2 below.

**Blocking**
1. GM->L1 Elem drops the `scratch` WRITE dependency (inline). `getEffects()` marks scratch as Write; the macro path bypasses `getEffects`, and the S phase has empty `defValues`, so scratch becomes an MTE2 read only -> un-synced RAW/WAW/WAR on a reused/aliased scratch buffer. One-line fix.

**Non-blocking**
2. Inference logic is duplicated in SyncMacroModel.cpp vs PTOToEmitC.cpp and must stay identical (drift -> wrong-pipe sync). Inline.
3. Dangling design-doc reference. Inline.
4. Validation gaps: the A5 1x1 scalar overload and the GM->L1 Elem scratch path are only checked via `--pto-insert-sync-debug=2` phase dumps / single-op lit, not by an emitted set_flag/wait_flag assertion and not board/camodel-verified. Also note the A5 1x1 model defs dst on PIPE_S while `getPipe()` reports PIPE_V for A5 GM->UB -- only the InsertSync path is corrected here; a getPipe-based solver (GraphSyncSolver) would still mis-pipe the 1x1 dst.

TaoTao-real:
静态看了一下这次 `pto.mgather` macro sync model 的适配，整体方向是对的：把 A2/A3 Row 形态建模成 `PIPE_S + PIPE_MTE2` 可以补上原来 `MGatherOp::getPipe()` 单 pipe 抽象隐藏掉的同步需求。不过建议合入前再补强几个边界：

1. **default coalesce 推断不要把 dynamic 直接当 Row 依据**

   `SyncMacroModel.cpp` 和 `PTOToEmitC.cpp` 里现在都把 `ShapedType::kDynamic` 当成 unit/compatible：
   - `SyncMacroModel.cpp` 的 `isUnit/compat`
   - `PTOToEmitC.cpp` 的 `isKnownUnitExtentForMGather/hasCompatibleKnownExtentForMGather`

   这在 verifier 里作为“动态形状暂时允许”是合理的，但在 EmitC/InsertSync 做静态选择时不够安全：如果 `valid_row/valid_col` 是运行时值，默认可能被推成 `Row`，但运行时实际可能是 Elem 形态。建议只有静态证明 index 是 `[0|1, R]` 或 `[R, 0|1]` 时才 infer Row；动态歧义场景最好要求显式 `coalesce`，或者保守选择不会改变语义的 lowering。

2. **zero-valid Row 与 verifier 规则不一致**

   `MGatherOp` verifier 明确允许 Row 形态为 `[0|1, dst.valid_row]` / `[dst.valid_row, 0|1]`，但这次新增的 coalesce inference 只认 `1` 和 dynamic，不认 `0`。这样合法的 zero-valid Row case 会被 EmitC 和 InsertSync 同时推成 Elem。建议把 unit 判断改成 `0 || 1`（同时结合上面 dynamic 的处理），并补一个 `v_row=0` / `v_col=0` 的 lit。

3. **GM->L1 Elem 的 scratch phase 建模漏了 def**

   注释说 Elem 路径里 scratch 是 “S def / MTE2 use”，但当前代码只在 `PIPE_MTE2` phase 的 use 里加入 scratch，`PIPE_S` phase 没有把 scratch 放进 defValues。由于 macro op 会走 `MakeMacroCompound`，只使用 phase 的 `defValues/useValues`，不再自动继承 op 的普通 `MemoryEffects`，这可能漏掉 scratch 对后续复用/读写的依赖。建议 Elem+scratch 时把 `op.getScratch()` 加到 S phase 的 defValues，同时保留 MTE2 use。

4. **A2/A3 hidden EVENT_ID0 reservation 建议给出 PTO-ISA 依据/测试**

   Row/Elem 分支现在对多个 pipe pair 都局部 reserve `EVENT_ID0`。这不是全局扣减 event id，但如果 reservation 比真实 macro 内部 fixed event 更多，会在 macro 周边制造额外 event pressure，可能触发不必要的 widen/barrier fallback。建议补充 A2/A3 Row/Elem 宏展开后的内部 set/wait 列表，或加 debug/FileCheck 覆盖 hidden reservation 对 event id 分配的影响。

5. **避免 EmitC 和 InsertSync 的 coalesce 推断逻辑漂移**

   现在两边各自实现了一份 shape/layout 恢复和 Row/Elem 判断。建议抽一个公共 helper；如果暂时不抽，至少补同一输入下 `MGATHER<Row/Elem>` 输出和 SyncMacroModel phase 一致的测试，尤其覆盖 dynamic/zero-valid/`bind_tile` 或 `pointer_cast` 元数据恢复场景。

我觉得这几个点里前 3 个属于合入前最好修掉的 correctness 边界；第 4/5 个是为了降低后续 sync/event-id 问题的定位成本。

reedhecre:
已接收 `/run a3`，A3 板测器会处理这条请求。

- 进度页：http://154.9.227.233/ptoas-board-dashboard/#board-a3
- 当前状态：板测器空闲，这条请求会在本轮轮询启动。

页面会自动刷新，可以直接看当前阶段、排队情况和最近结果。

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`manual`
- 源码提交：`9a60d52e33df`
- 结果汇总：OK `221` / FAIL `0` / SKIP `1`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260629_091410_manual_pr865.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260629_091410_manual_pr865.tsv`
- 手动指令：`/run a3`
- 触发人：`FangRui0`
- 触发评论：https://github.com/hw-native-sys/PTOAS/pull/865#issuecomment-4828161379

lib/PTO/IR/PTO.cpp
```diff
@@ -8499,12 +8499,14 @@ LogicalResult MGatherOp::verify() {
                                              "dst")))
     return failure();
 
-  std::optional<pto::Coalesce> explicitCoalesce;
-  if (auto coalesceAttr = getCoalesceAttr())
-    explicitCoalesce = coalesceAttr.getValue();
+  auto coalesceAttr = getCoalesceAttr();
+  if (!coalesceAttr)
```
zhangstevenunity, line 8503: Hard-requiring `coalesce` here (and the mirror error added in `PTOMGatherToMGATHER`) regresses the public contract: `coalesce` is still `OptionalAttr` in ODS (`PTOOps.td:2858`) and the manual still documents omission→inference for GM→UB (`PTO_IR_manual.md:7472`). It also breaks an in-tree producer — `test/samples/Mgather/mgather.py:62` emits an `mgather` with no coalesce and then calls `verify()` — which this PR does not patch.

If explicit coalesce is the intended contract, make it so across all layers (ODS required/DefaultValued, Python builder, the sample, and the manual). Otherwise keep inference for GM→UB and require explicit only where there is no UB index to infer from (GM→L1) or where the index shape is dynamically ambiguous.

Note the memref-typed path skips this op verifier and only trips the identical error later in EmitC — worth unifying so the failure surface is consistent.

lib/PTO/Transforms/PTOToEmitC.cpp
```diff
@@ -380,12 +380,12 @@ static Value maybeWrapGlobalMemrefAsGlobalTensor(
...
 
 static bool hasCompatibleKnownExtentForMGather(int64_t lhs, int64_t rhs) {
-  return lhs == ShapedType::kDynamic || rhs == ShapedType::kDynamic ||
+  return lhs != ShapedType::kDynamic && rhs != ShapedType::kDynamic &&
          lhs == rhs;
 }
 
 static bool isKnownUnitExtentForMGather(int64_t value) {
-  return value == ShapedType::kDynamic || value == 1;
+  return value == 0 || value == 1;
```
zhangstevenunity, line 388: After this PR these two helpers are only reached by MScatter coalesce inference (line ~6240); MGather no longer infers. The semantics changed in the same edit: `isKnownUnitExtentForMGather` now accepts `0` (good — matches the verifier's `[0|1, R]` Row form) but no longer treats `kDynamic` as unit, and `hasCompatibleKnownExtentForMGather` now rejects dynamic instead of accepting it.

Consequence: a dynamic-shaped MScatter that previously inferred `Row` will now infer `Elem` — a silent behavior change with no test. Either apply the same explicit-coalesce policy to MScatter for consistency, or add a dynamic / zero-valid MScatter lit to lock the new inference behavior down.

zhangstevenunity (changes requested):
Re-reviewed at head `d37fd4d3` (was `91fcde7a`). The blocking scratch bug from my last pass is **fixed** — but the new "require explicit coalesce" commit introduces a different, incomplete breaking change.

**Resolved since `91fcde7a`**
- GM→L1 Elem scratch is now modeled as an `S` def + `MTE2` use, matching `MGatherOp::getEffects` (which already marks scratch `Write`, PTO.cpp:13102-13104), with a reuse regression test. The macro-model core — the actual #861 fix — is sound on every path (A2/A3 Row/Elem, A5 SIMT, A5 1x1 scalar, GM→L1).

**Blocking (new): `d37fd4d3` makes `coalesce` hard-required, but only in C++**
Same regression the Codex bot flags, and it is concrete:
- ODS still declares `coalesce` `OptionalAttr` (`include/PTO/IR/PTOOps.td:2858`); the manual still documents omission→inference for GM→UB (`docs/PTO_IR_manual.md:7445`, `:7472`). Neither is updated, so the op's stated contract now disagrees with its own verifier.
- In-tree producer breaks: `test/samples/Mgather/mgather.py:62` builds `pto.MGatherOp(sv0, tb1, tb2)` with no coalesce and calls `m.operation.verify()` at `:69` → now hard-fails. The PR patches one lit test (`mgather_mscatter_low_precision_a5_emitc.pto`) but not this sample.
- The A3 board run (OK `221` / FAIL `0`, source `9a60d52e`) was triggered at `01:13` on 06-29; this commit landed at `10:29` — so the require-coalesce change is **not** board-validated.

Either complete the cross-layer change (ODS required/DefaultValued + Python builder + sample + docs) or narrow it (keep inference for GM→UB, require explicit only for GM→L1 and dynamically-ambiguous cases). Inline detail below.

**Non-blocking**
- Asymmetry / silent behavior change: inference is removed for MGather but **kept for MScatter** (`PTOToEmitC.cpp:6240`), and this PR changes the shared helpers. A genuinely dynamic-shaped MScatter that previously inferred `Row` now infers `Elem`, with no test. If inference is unsafe enough to ban for MGather, it is unsafe for MScatter too.
- The `0`-unit handling is a real improvement (aligns with the verifier's `[0|1, R]` Row form) but is now only reachable via MScatter; no zero-valid MScatter lit added.
- `lookupValidDims` / `lookupTileConfig` / `getConstantIndex` are duplicated near-verbatim between `SyncMacroModel.cpp` and `PTOToEmitC.cpp`.

Net: the mechanism is correct and my prior blocker is resolved; the remaining blocker is the incomplete/unvalidated require-coalesce change.

lib/PTO/Transforms/PTOToEmitC.cpp
zhangstevenunity, line 388: Non-blocking: after this PR these helpers are reached only by MScatter inference (line ~6240) — MGather now requires explicit coalesce. Tightening `isKnownUnitExtentForMGather` (now `0 || 1`, no longer `kDynamic`) and `hasCompatibleKnownExtentForMGather` (now rejects `kDynamic`) silently changes MScatter: an idx with runtime-dynamic valid dims that previously inferred `Row` now infers `Elem` (different scatter semantics and sync). Consider applying the same explicit-coalesce contract to MScatter, or adding a dynamic / zero-valid MScatter lit to lock this down.

zhangstevenunity (approved):
Re-reviewed at head `ab5ea4ab`. **Blocker resolved — approving.**

The new commit `ab5ea4ab` ("align mgather explicit coalesce contract") completes the cross-layer change I flagged:

- **ODS**: `coalesce` is now required (`PTO_CoalesceAttr:$coalesce`, `PTOOps.td:2858`) — the op contract finally matches the verifier.
- **Docs**: manual updated (default `inferred`→`required`, semantics, examples) (`PTO_IR_manual.md:7426-7497`).
- **Sample**: `test/samples/Mgather/mgather.py:62` now passes `coalesce=row`, which is exactly what the old inference selected for its `[1,32]` idx → `[32,32]` dst, so behavior is preserved.
- **Lit**: `mgather_default_coalesce_static_only.pto` now checks the ODS-generated `requires attribute 'coalesce'`.
- `PTOViewToMemref` forwards `getCoalesceAttr()` (`PTOViewToMemref.cpp:3891`), so the required attr survives the tile→memref rewrite and the macro model / `getPipe()` see it.
- CI is green on both arches + `build-and-test` + lit; the npu-validation generator only string-matches `"MGATHER"` in emitted kernels (not an IR producer), so it is unaffected.

The macro-sync model itself (the actual #861 fix) was already sound and board-validated (A3 221/0), with the GM→L1 Elem `scratch` dependency now correctly modeled as an `S` def + `MTE2` use.

**Non-blocking follow-ups (not gating):**
1. Asymmetry: MScatter still infers coalesce (`PTOOps.td:3069` `OptionalAttr`) using the helpers this PR tightened — see inline. A genuinely dynamic-shaped MScatter that previously inferred `Row` now infers `Elem`. If inference is unsafe enough to ban for gather, the same applies to scatter; worth a follow-up to either require it there too or add a dynamic-shape MScatter lit.
2. Dead code: now that ODS enforces presence, the manual `if (!coalesceAttr)` checks in `MGatherOp::verify()` and `getMGatherSyncMacroModel` are unreachable — safe to drop.
3. The sample fix is correct by inspection but is not exercised by CI (`mgather.py` is `target_arch=a5`, which the default a3 sample lane skips).

Thanks for the quick turnaround on the contract alignment.

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`f105bfa613eb`
- 结果汇总：OK `221` / FAIL `0` / SKIP `1`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260630_170120_merged_pr865.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260630_170120_merged_pr865.tsv`

Issue #870, opened 2026-06-27
[Bug] ptoas codegen generates dead TMOV self-copy and redundant pipe_barrier in AIV v_proj_vec_rs kernel


Inspiron-st:
### Diagnosis

**ptoas** — The generated kernel file header says "Generated by PyPTO IR Compiler (PTO backend)" / "ptoas-generated code". The dead TMOV self-copy and redundant pipe_barrier(PIPE_V) are emitted by ptoas's codegen pass.

### Description

ptoas generates dead `TMOV` self-copy instructions and redundant `pipe_barrier(PIPE_V)` in the AIV `v_proj_vec_rs` kernel.

**Reproduction:**
1. Build a model with `V_PROJ_ON_AIV=1` (e.g. Qwen3-14B decode layer)
2. Run `python decode_layer.py -p a2a3 -d 0 --fwd-layers 2`
3. Inspect the generated kernel at `build_output/_jit_decode_fwd_*/kernels/aiv/v_proj_vec_rs.cpp`

**Issue 1: Dead TMOV self-copy (320 instances)**

Inside the inner N-column loop (executed 160 times × 2 groups = 320 total), ptoas emits:

```cpp
// Line 165-169 (first group), repeated symmetrically at line 226-230
Tile<..., 16, 16> v77 = ...;
TASSIGN(v77, v78);           // v78 = 65600 (v22)
pipe_barrier(PIPE_V);
TMOV(v77, v28);              // v28@65600 → v77@65600 — SELF-COPY, v77 never read again
```

`v28` (the accumulator) is assigned to UB address 65600 (line 76, `TASSIGN(v28, v22=65600)`). `v77` is assigned to the **same address** 65600 (`v78 = v22 = 65600`). `TMOV(v77, v28)` copies 65600→65600 (a self-copy), and `v77` is never read by any subsequent instruction — the next loop iteration redeclares it.

This produces 320 dead `TMOV` instructions, each accompanied by:
- 1 redundant `Tile` declaration
- 1 redundant `TASSIGN`
- 1 redundant `pipe_barrier(PIPE_V)`

Total dead code: ~960 lines of useless instructions + 320 useless memory copies.

**Issue 2: Redundant pipe_barrier(PIPE_V) (~1001 instances)**

[...]

PR #420, opened 2026-04-02
[A5][Sync] remove identity tmov before insert-sync

TaoTao-real:
Summary
- Add an A5-only cleanup pass (`pto-remove-identity-tmov`) that erases `pto.tmov ins(%x) outs(%x)` before auto-sync runs.
- Wire the pass in `ptoas` directly before `PTOInsertSync` when `--enable-insert-sync` is enabled.
- Add regression test `test/basic/identity_tmov_autosync_a5_only.pto`.

Motivation
- Fixes the A5 hang risk caused by identity `tmov` being treated as a real producer/consumer by auto-sync, which can create spurious sync edges around a hardware no-op move.

Design
- New pass: `PTORemoveIdentityTMovPass` (func pass).
- Gated by module attribute `pto.target_arch == "a5"`.
- Removes only must-prove identity `tmov` (`src == dst` SSA value).
- If optional result is used and type-compatible with `dst`, rewires uses to `dst` before erase.
- Pipeline placement: `PTOResolveReservedBuffers -> PTORemoveIdentityTMov -> PTOInsertSync`.

Testing
- Built: `ninja -C /Users/lishengtao/Documents/PTO/_codex_worktrees/ptoas_identity_tmov_a5/build ptoas`
- New targeted checks:
  - `ptoas --pto-arch=a5 --enable-insert-sync test/basic/identity_tmov_autosync_a5_only.pto | FileCheck ... --check-prefix=A5`
  - `ptoas --pto-arch=a3 --enable-insert-sync test/basic/identity_tmov_autosync_a5_only.pto | FileCheck ... --check-prefix=A3`
- Extra guard:
  - `ptoas --pto-arch=a5 --enable-insert-sync test/basic/tmov_acc_mat_pipe_selection.pto | FileCheck ...`

Risk / Rollback
- Risk is low: behavior change is strictly A5-scoped and only for syntactic identity moves.
- Rollback is straightforward: revert this PR.

PR #876, opened 2026-06-29
fix(sync): prune identity tmov before insert sync

TaoTao-real:
Summary
- Add a pre-InsertSync `pto-remove-identity-tmov` cleanup pass.
- Remove plain `pto.tmov` when src/dst are proven to refer to the same physical tile address/range.
- Place the pass after reserved-buffer resolution and before automatic sync insertion so sync is not generated for the no-op TMOV.
- Add regression coverage that checks identity TMOV and its induced `pipe_barrier(PIPE_V)` disappear, while different-address and unproven-dynamic TMOV are preserved.

Motivation
- Fixes the first part of #870: dead identity TMOV self-copy and the redundant sync generated only because that no-op still reached InsertSync.
- Supersedes and absorbs the useful part of the earlier identity-TMOV cleanup attempts #419/#420:
  - #419 is closed and was not merged.
  - #420 is still open and was not merged into current `main`.
  - #420's useful exact-address-range proof is incorporated here, but without its broader PR scope/debug flag/unrelated diagnostic cases.
- This PR intentionally does not change generic same-pipe `PIPE_V` barrier policy; that will be handled separately.

Design
- The cleanup is conservative and only removes plain TMOV:
  - no `fp`
  - no `preQuantScalar`
  - no `accToVecMode`
  - `reluPreMode == NoRelu`
  - src/dst/result types match
- Address identity is proven in two tiers:
  - fast explicit address proof: direct same SSA, `pto.bind_tile` / `memref.cast` peeling, `pto.tassign`, and single-address `pto.pointer_cast`
  - fallback exact-range proof: lazily build InsertSync's `Buffer2MemInfoMap` and require same scope, same non-zero allocate size, identical base addresses, and either the same InsertSync root or a different-root same-concrete-address TMOV whose destination is dead after the TMOV
- Dynamic or ambiguous cases are kept. If exact identity cannot be proven, the TMOV remains. Different-root same-address TMOVs are also kept when the destination is live, because the TMOV may be the only bridge that lets InsertSync observe the src-root to dst-root dependency.

Performance Profiling / camodel-msprof
- Actual collection was performed on the A3 remote host through `task-submit + msprof op simulator`.
- Because the full Qwen `v_proj_vec_rs.cpp` artifact from #870 was not available in the current run environment, this measures a focused AIV/vector microbenchmark that reproduces the exact identity-TMOV pattern fixed by this PR.
- Microbenchmark shape:
  - common work in both variants: `TLOAD -> pipe_barrier(PIPE_ALL) -> TADD -> pipe_barrier(PIPE_ALL) -> TSTORE`
  - baseline-only redundant work: 320 iterations of `pipe_barrier(PIPE_V); TMOV(alias, acc)` where `alias` and `acc` are `TileType::Vec<f32, 32x32>` bound to the same UB address (`0x1000`)
  - candidate: same kernel with the identity-TMOV pattern removed
  - correctness check in both runs: `repeat=320 expected=2 first=2 max_abs_err=0`
- Environment:
  - CANN: `/usr/local/Ascend/cann-9.0.0`
  - simulator SOC: `dav_2201`
  - pto-isa headers: `/home/zhongxuan/pto-isa`
  - workdir: `/data/pyptouser/zhongxuan/ptoas_board_runs/issue870_identity_tmov_camodel`
- Collection command:
  ```bash
  task-submit --device auto --max-time 600 --timeout 60 --run     "bash /data/pyptouser/zhongxuan/ptoas_board_runs/issue870_identity_tmov_camodel/run_msprof_task.sh <baseline|candidate> 320"
  ```
- Results:

  | Variant | core0.veccore0 duration | core0.veccore0 running | Total tick | instr rows | redundant `BAR PIPE:VEC` | redundant `MOV_UB_TO_UB` | Verification |
  |---|---:|---:|---:|---:|---:|---:|---|
  | baseline | 3.56 us | 3.56 us | 8123 | 75 | 320 | 320 | pass |
  | candidate | 0.62 us | 0.62 us | 2672 | 60 | 0 | 0 | pass |

- Delta on this focused pattern:
  - veccore duration: `3.56us -> 0.62us`, `-2.94us` (`-82.6%`)
  - total simulator tick: `8123 -> 2672`, `-5451` (`-67.1%`)
- CSV evidence:
  - baseline `core0.veccore0_instr_exe.csv` contains `BAR, pipe=VECTOR, call_count=320, detail=PIPE:VEC` and `MOV_UB_TO_UB, call_count=320, Src:UB,Dst:UB,XD:X4=0x1000,XN:X4=0x1000`
  - candidate `core0.veccore0_instr_exe.csv` has neither redundant row
- Scope note: this is not a full Qwen decode end-to-end number. It directly validates the #870 identity-TMOV pattern removed by this PR and the induced `PIPE_V` barrier disappearing before InsertSync/codegen reaches profiling.

Testing
- Local license check:
  - `python3 .github/scripts/check_license_headers.py --repo hw-native-sys/PTOAS --event-name push --base-sha hw-native-sys/main --head-sha HEAD` OK
- Local targeted compile check:
  - `ninja -C build-llvm21 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o` OK
- Follow-up guard after CI signal:
  - Low-precision `TMOV` operands/results (`hif8`/`fp4` family through `isPTOLowPrecisionType`) are explicitly preserved by the identity cleanup pass.
  - This keeps existing A5 low-precision EmitC coverage, including `tmov_low_precision_a5_emitc.pto` and `tmov_fp4_low_precision_a5_emitc.pto`, from being optimized away.
- P1 review follow-up:
  - Different-root same-address TMOVs are now pruned only when the destination has no uses outside the TMOV and all TMOV results are unused; live destination cases keep the TMOV to preserve the InsertSync bridge and avoid under-synchronization.
  - Regression coverage added for a live-dst same-address case that must keep `TMOV`.
  - Integer cast chains in explicit address proofs are now evaluated with exact `sext`/`zext`/`trunc` semantics instead of being stripped transparently.
  - Regression coverage added for `extui(i8 -1) -> i64` vs `i64 -1`, and `trunci(i64 255) -> i8` followed by `extsi -> i64` vs `i64 255`.
- Previous local LLVM21 validation worktree with the base patch:
  - `ninja -C build-llvm21 ptoas` OK
  - `llvm-lit -v build-llvm21/test/lit/pto/issue870_identity_tmov_pruning.pto` OK
  - `llvm-lit -v build-llvm21/test/lit/pto/issue454_nested_loop_same_pipe_pair_regression_gss.pto build-llvm21/test/lit/pto/syncfinder_zero_loop_if_probe_gss.pto build-llvm21/test/lit/pto/issue706_tgather_hidden_mte_sync.pto build-llvm21/test/lit/pto/issue706_tscatter_hidden_mte_sync.pto` OK
- Note: direct full rebuild of the fresh upstream-main worktree on this machine is currently blocked by a pre-existing local LLVM header/API mismatch in unrelated PTO IR files (`Type::isFloat8*`, `getStridesAndOffset`). The new pass object itself compiles cleanly in that environment.

Risk / Rollback
- Risk is low because the pass only erases TMOV when it can prove physical self-copy and the TMOV has no extra conversion/quant/relu semantics. Low-precision TMOV is excluded from cleanup and remains on the existing lowering path. Integer address casts are modeled exactly; unmodeled cast chains remain conservative and do not trigger deletion.
- Rollback: revert this PR; no public IR, CLI, or API changes are required.

### DIFF
--- a/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
+++ b/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
@@ -36,11 +36,6 @@ using namespace mlir::pto;
 
 namespace {
 
-struct AddressToken {
-  Value value;
-  std::optional<int64_t> constant;
-};
-
 static std::optional<unsigned> getIntegerLikeBitWidth(Type type) {
   if (auto intTy = dyn_cast<IntegerType>(type))
     return intTy.getWidth();
@@ -110,43 +105,6 @@ static std::optional<int64_t> tryEvalI64Constant(Value value) {
   if (!apInt || apInt->getBitWidth() > 64)
     return std::nullopt;
   return apInt->getSExtValue();
-}
-
-static Value peelMetadata(Value value) {
-  while (true) {
-    if (auto bind = value.getDefiningOp<BindTileOp>()) {
-      value = bind.getSource();
-      continue;
-    }
-    if (auto cast = value.getDefiningOp<memref::CastOp>()) {
-      value = cast.getSource();
-      continue;
-    }
-    return value;
-  }
-}
-
-static std::optional<AddressToken> getExplicitAddressToken(Value value) {
-  value = peelMetadata(value);
-
-  if (auto tassign = value.getDefiningOp<TAssignOp>())
-    return AddressToken{tassign.getAddr(),
-                        tryEvalI64Constant(tassign.getAddr())};
-
-  auto pointerCast = value.getDefiningOp<PointerCastOp>();
-  if (!pointerCast || pointerCast.getAddrs().size() != 1)
-    return std::nullopt;
-
-  Value addr = pointerCast.getAddrs().front();
-  return AddressToken{addr, tryEvalI64Constant(addr)};
-}
-
-static bool sameAddressToken(const AddressToken &lhs, const AddressToken &rhs) {
-  if (lhs.value == rhs.value)
-    return true;
-  if (lhs.constant && rhs.constant)
-    return *lhs.constant == *rhs.constant;
-  return false;
 }
 
 static bool hasOnlyCurrentOpUses(Value value, Operation *currentOp) {
@@ -262,6 +220,56 @@ static bool hasExactSameAddressRange(const BaseMemInfo *srcInfo,
   return true;
 }
 
+static bool hasSameConcreteAddressRange(const BaseMemInfo *srcInfo,
+                                        const BaseMemInfo *dstInfo) {
+  if (!hasExactSameAddressRange(srcInfo, dstInfo))
+    return false;
+  if (srcInfo->rootBuffer == dstInfo->rootBuffer)
+    return true;
+  auto srcRootAddr = tryGetConcreteRootAddress(srcInfo);
+  auto dstRootAddr = tryGetConcreteRootAddress(dstInfo);
+  return srcRootAddr && dstRootAddr && *srcRootAddr == *dstRootAddr;
+}
+
+static Operation *getAncestorInBlock(Operation *op, Block *block) {
+  for (Operation *cur = op; cur; cur = cur->getParentOp()) {
+    if (cur->getBlock() == block)
+      return cur;
+  }
+  return nullptr;
+}
+
+static bool hasUseAfterOp(Value value, Operation *currentOp) {
+  Block *block = currentOp->getBlock();
+  for (OpOperand &use : value.getUses()) {
+    Operation *owner = use.getOwner();
+    if (owner == currentOp)
+      continue;
+    Operation *ancestor = getAncestorInBlock(owner, block);
+    if (!ancestor)
+      return true;
+    if (ancestor != currentOp && currentOp->isBeforeInBlock(ancestor))
+      return true;
+  }
+  return false;
+}
+
+static bool hasLaterUseOfSameAddressRange(
+    TMovOp op, const BaseMemInfo *dstInfo,
+    const Buffer2MemInfoMap &buffer2MemInfoMap) {
+  for (const auto &entry : buffer2MemInfoMap) {
+    // Later reads of the source itself do not make this no-op TMOV a bridge.
+    if (entry.first == op.getSrc())
+      continue;
+    bool sameRange = llvm::any_of(entry.second, [&](const auto &info) {
+      return hasSameConcreteAddressRange(info.get(), dstInfo);
+    });
+    if (sameRange && hasUseAfterOp(entry.first, op))
+      return true;
+  }
+  return false;
+}
+
 static bool hasPlainTMovSemantics(TMovOp op) {
   return !op.getFp() && !op.getPreQuantScalar() && !op.getAccToVecModeAttr() &&
          op.getReluPreMode() == ReluPreMode::NoRelu;
@@ -297,17 +305,6 @@ static bool touchesLowPrecisionElement(TMovOp op) {
   });
 }
 
-static bool isIdentityTMovByExplicitAddress(TMovOp op) {
-  if (op.getSrc() == op.getDst())
-    return true;
-
-  std::optional<AddressToken> src = getExplicitAddressToken(op.getSrc());
-  std::optional<AddressToken> dst = getExplicitAddressToken(op.getDst());
-  if (!src || !dst)
-    return false;
-  return sameAddressToken(*src, *dst) && isDeadDstTMov(op);
-}
-
 static bool
 isIdentityTMovByMemInfo(TMovOp op, const Buffer2MemInfoMap &buffer2MemInfoMap) {
   Value src = op.getSrc();
@@ -318,16 +315,11 @@ isIdentityTMovByMemInfo(TMovOp op, const Buffer2MemInfoMap &buffer2MemInfoMap) {
 
   const BaseMemInfo *srcInfo = getSingleMemInfo(buffer2MemInfoMap, src);
   const BaseMemInfo *dstInfo = getSingleMemInfo(buffer2MemInfoMap, dst);
-  if (!hasExactSameAddressRange(srcInfo, dstInfo))
+  if (!hasSameConcreteAddressRange(srcInfo, dstInfo))
     return false;
 
-  if (srcInfo->rootBuffer == dstInfo->rootBuffer)
-    return true;
-
-  auto srcRootAddr = tryGetConcreteRootAddress(srcInfo);
-  auto dstRootAddr = tryGetConcreteRootAddress(dstInfo);
-  return srcRootAddr && dstRootAddr && *srcRootAddr == *dstRootAddr &&
-         isDeadDstTMov(op);
+  return isDeadDstTMov(op) &&
+         !hasLaterUseOfSameAddressRange(op, dstInfo, buffer2MemInfoMap);
 }
 
 struct PTORemoveIdentityTMovPass
@@ -342,7 +334,7 @@ struct PTORemoveIdentityTMovPass
       if (!hasPlainTMovSemantics(op) || !hasCompatibleIdentityTypes(op) ||
           touchesLowPrecisionElement(op))
         return;
-      if (isIdentityTMovByExplicitAddress(op)) {
+      if (op.getSrc() == op.getDst()) {
         identityMoves.push_back(op);
         return;
       }
--- a/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
+++ b/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
@@ -1,0 +1,134 @@
+// Copyright (c) 2026 Huawei Technologies Co., Ltd.
+// This program is free software, you can redistribute it and/or modify it under
+// the terms and conditions of CANN Open Software License Agreement Version 2.0
+// (the "License"). Please refer to the License for details. You may not use
+// this file except in compliance with the License. THIS SOFTWARE IS PROVIDED ON
+// AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS
+// FOR A PARTICULAR PURPOSE. See LICENSE in the root of the software repository
+// for the full text of the License.
+
+//===- PTORemoveIdentityTMov.cpp -----------------------------------------===//
+//===----------------------------------------------------------------------===//
+
+#include "PTO/IR/PTO.h"
+#include "PTO/Transforms/Passes.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/Pass/Pass.h"
+
+#include <optional>
+
+namespace mlir {
+namespace pto {
+#define GEN_PASS_DEF_PTOREMOVEIDENTITYTMOV
+#include "PTO/Transforms/Passes.h.inc"
+} // namespace pto
+} // namespace mlir
+
+using namespace mlir;
+using namespace mlir::pto;
+
+namespace {
+
+struct AddressToken {
+  Value value;
+  std::optional<llvm::APInt> constant;
+};
+
+static Value peelMetadata(Value value) {
+  while (true) {
+    if (auto bind = value.getDefiningOp<BindTileOp>()) {
+      value = bind.getSource();
+      continue;
+    }
+    if (auto cast = value.getDefiningOp<memref::CastOp>()) {
+      value = cast.getSource();
+      continue;
+    }
+    return value;
+  }
+}
+
+static std::optional<AddressToken> getExplicitAddressToken(Value value) {
+  value = peelMetadata(value);
+
+  if (auto tassign = value.getDefiningOp<TAssignOp>())
+    return AddressToken{tassign.getAddr(), std::nullopt};
+
+  auto pointerCast = value.getDefiningOp<PointerCastOp>();
+  if (!pointerCast || pointerCast.getAddrs().size() != 1)
+    return std::nullopt;
+
+  Value addr = pointerCast.getAddrs().front();
+  llvm::APInt constant;
+  if (matchPattern(addr, m_ConstantInt(&constant)))
+    return AddressToken{addr, constant};
+  return AddressToken{addr, std::nullopt};
+}
+
+static bool sameAddressToken(const AddressToken &lhs, const AddressToken &rhs) {
+  if (lhs.value == rhs.value)
+    return true;
+  if (lhs.constant && rhs.constant)
+    return lhs.constant->getSExtValue() == rhs.constant->getSExtValue();
+  return false;
+}
+
+static bool hasPlainTMovSemantics(TMovOp op) {
+  return !op.getFp() && !op.getPreQuantScalar() && !op.getAccToVecModeAttr() &&
+         op.getReluPreMode() == ReluPreMode::NoRelu;
+}
+
+static bool hasCompatibleIdentityTypes(TMovOp op) {
+  if (op.getSrc().getType() != op.getDst().getType())
+    return false;
+  for (OpResult result : op->getResults()) {
+    if (result.getType() != op.getDst().getType())
+      return false;
+  }
+  return true;
+}
+
+static bool isIdentityTMov(TMovOp op) {
+  if (!hasPlainTMovSemantics(op) || !hasCompatibleIdentityTypes(op))
+    return false;
+
+  if (op.getSrc() == op.getDst())
+    return true;
+
+  std::optional<AddressToken> src = getExplicitAddressToken(op.getSrc());
+  std::optional<AddressToken> dst = getExplicitAddressToken(op.getDst());
+  if (!src || !dst)
+    return false;
+  return sameAddressToken(*src, *dst);
+}
+
+struct PTORemoveIdentityTMovPass
+    : public mlir::pto::impl::PTORemoveIdentityTMovBase<
+          PTORemoveIdentityTMovPass> {
+  void runOnOperation() override {
+    func::FuncOp func = getOperation();
+    SmallVector<TMovOp, 16> identityMoves;
+
+    func.walk([&](TMovOp op) {
+      if (isIdentityTMov(op))
+        identityMoves.push_back(op);
+    });
+
+    for (TMovOp op : identityMoves) {
+      for (OpResult result : op->getResults())
+        result.replaceAllUsesWith(op.getDst());
+      op.erase();
+    }
+  }
+};
+
+} // namespace
+
+std::unique_ptr<Pass> mlir::pto::createPTORemoveIdentityTMovPass() {
+  return std::make_unique<PTORemoveIdentityTMovPass>();
+}
--- a/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
+++ b/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
@@ -1,12 +1,10 @@
 // Copyright (c) 2026 Huawei Technologies Co., Ltd.
-// This program is free software, you can redistribute it and/or modify it under
-// the terms and conditions of CANN Open Software License Agreement Version 2.0
-// (the "License"). Please refer to the License for details. You may not use
-// this file except in compliance with the License. THIS SOFTWARE IS PROVIDED ON
-// AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
-// INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS
-// FOR A PARTICULAR PURPOSE. See LICENSE in the root of the software repository
-// for the full text of the License.
+// This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+// CANN Open Software License Agreement Version 2.0 (the "License").
+// Please refer to the License for details. You may not use this file except in compliance with the License.
+// THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+// See LICENSE in the root of the software repository for the full text of the License.
 
 //===- PTORemoveIdentityTMov.cpp -----------------------------------------===//
 //===----------------------------------------------------------------------===//
--- a/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
+++ b/lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
@@ -10,6 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "PTO/IR/PTO.h"
+#include "PTO/IR/PTOTypeUtils.h"
 #include "PTO/Transforms/InsertSync/MemoryDependentAnalyzer.h"
 #include "PTO/Transforms/InsertSync/PTOIRTranslator.h"
 #include "PTO/Transforms/InsertSync/SyncCommon.h"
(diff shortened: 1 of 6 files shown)


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #876 fix(sync): prune identity tmov before insert sync
- Author: `TaoTao-real`
- Base/Head: `main` / `codex/issue870-identity-tmov`
- Head SHA: `1ed680d30caa`
- Trigger: PR 有新提交
- Generated At: `2026-07-01T04:51:12Z`
- Previous Head SHA: `0a0a545cb7be`
- Status: failed at `codex-review` (timeout=1800s)

### Summary

Review failed at stage `codex-review`: timeout=1800s

### Findings

### Log Tail

```text
#define GEN_PASS_DEF_PTOA5NORMALIZETMOV
#include "PTO/Transforms/Passes.h.inc"
} // namespace pto
} // namespace mlir
exec
/bin/bash -lc "grep -RIn \"pto.treshape\\|view_semantics = \\\"treshape\\\"\\|A5NormalizeTMov\" test/lit/pto lib/PTO/Transforms | head -n 200" in /tmp/ptoas-pr-review-monitor/runs/20260701_122028_pr876/repo succeeded in 0ms:
test/lit/pto/treshape_static_valid_shape_emitc.pto:12:    %reshaped = pto.treshape %src
test/lit/pto/treshape_explicit_dynamic_valid_shape_preserved.pto:15:    %dst = pto.treshape %src
test/lit/pto/treshape_explicit_dynamic_valid_shape_preserved.pto:38:// CHECK: pto.bind_tile {{.*}}, [[PART]], [[FULL]] {{.*}}blayout=#pto.blayout<col_major>{{.*}}pto.view_semantics = "treshape"
test/lit/pto/issue686_a5_tmov_treshape_square_dynamic_valid_shape.pto:34:// CHECK: [[SRC:%[0-9]+]] = pto.bind_tile [[SRC_BASE]], [[PART]], [[FULL]] {{.*}}blayout=#pto.blayout<row_major>{{.*}}pto.view_semantics = "treshape"
test/lit/pto/issue686_a5_tmov_treshape_square_dynamic_valid_shape.pto:35:// CHECK: [[DST:%[0-9]+]] = pto.bind_tile [[DST_BASE]], [[PART]], [[FULL]] {{.*}}blayout=#pto.blayout<row_major>{{.*}}pto.view_semantics = "treshape"
lib/PTO/Transforms/GraphSyncSolver/PTOGraphSyncSolver.cpp:70:    // source as LoweringSyncToPipe / PTOA5NormalizeTMov / PTOPlanMemory).
lib/PTO/Transforms/PTOA5NormalizeTMovPass.cpp:171:struct PTOA5NormalizeTMovPass
lib/PTO/Transforms/PTOA5NormalizeTMovPass.cpp:172:    : public mlir::pto::impl::PTOA5NormalizeTMovBase<PTOA5NormalizeTMovPass> {
lib/PTO/Transforms/PTOA5NormalizeTMovPass.cpp:245:          "expected normalization to row_major via pto.treshape");
lib/PTO/Transforms/PTOA5NormalizeTMovPass.cpp:256:std::unique_ptr<Pass> mlir::pto::createPTOA5NormalizeTMovPass() {
lib/PTO/Transforms/PTOA5NormalizeTMovPass.cpp:257:  return std::make_unique<PTOA5NormalizeTMovPass>();
lib/PTO/Transforms/CMakeLists.txt:58:  PTOA5NormalizeTMovPass.cpp
lib/PTO/Transforms/FoldTileBufIntrinsics.cpp:188:          "FoldTileBufIntrinsics: pto.treshape must produce !pto.tile_buf");
lib/PTO/Transforms/FoldTileBufIntrinsics.cpp:202:                  "(pto.alloc_tile, pto.materialize_tile, or pto.treshape, "
exec
/bin/bash -lc "grep -n \"TReshapeOp\\|view_semantics = \\\"treshape\\\"\\|create<pto::BindTileOp>\" lib/PTO/Transforms/PTOViewToMemref.cpp | head -n 120" in /tmp/ptoas-pr-review-monitor/runs/20260701_122028_pr876/repo succeeded in 0ms:
964:      auto bindOp = rewriter.create<pto::BindTileOp>(
976:    auto bindOp = rewriter.create<pto::BindTileOp>(
1016:    auto bindOp = rewriter.create<pto::BindTileOp>(
1486:    auto bindOp = rewriter.create<pto::BindTileOp>(
1536:  auto bindOp = rewriter.create<pto::BindTileOp>(
1546:  DefaultInlineVector<mlir::pto::TReshapeOp> reshapes;
1547:  func.walk([&](mlir::pto::TReshapeOp op) { reshapes.push_back(op); });
1672:          auto bindOp = rewriter.create<pto::BindTileOp>(
1751:          auto bindOp = rewriter.create<pto::BindTileOp>(
1766:        auto bindOp = rewriter.create<pto::BindTileOp>(
1810:        auto bindOp = rewriter.create<pto::BindTileOp>(
exec
/bin/bash -lc "sed -n '11780,12430p' lib/PTO/Transforms/PTOToEmitC.cpp" in /tmp/ptoas-pr-review-monitor/runs/20260701_122028_pr876/repo succeeded in 0ms:
  };

  enum class MxValidAlignPolicy {
    None,
    AlignRowToShape,
    AlignColToShape,
  };

  static bool getIndexConst(Value v, int64_t &out) {
    if (!v)
      return false;
    if (auto cst = v.getDefiningOp<arith::ConstantOp>()) {
      if (auto ia = dyn_cast<IntegerAttr>(cst.getValue())) {
        out = ia.getValue().getSExtValue();
        return true;
      }
    }
    return false;
  }

exec
/bin/bash -lc "sed -n '210,280p' lib/PTO/Transforms/PTOMaterializeTileHandles.cpp" in /tmp/ptoas-pr-review-monitor/runs/20260701_122028_pr876/repo succeeded in 0ms:

  if (isa<TMatmulAccOp>(owner)) {
    if (!isA5Target(owner))
      return;
    if (operandNo == 0 || operandNo == 1 || operandNo == 3)
      meta.config = colRow();
    else if (operandNo == 2)
      meta.config = rowCol();
    return;
  }

  if (isa<TInsertOp>(owner)) {
    if (operandNo != 0 && operandNo != 3)
      return;
    auto as = getAddressSpace(operandType);
    if (!as)
      return;
    if (*as == AddressSpace::ACC || *as == AddressSpace::MAT)
      meta.config = colRow();
  }
exec
[monitor] stage codex-review exceeded timeout 1800s, terminating process group
===== END STAGE codex-review rc=-15 timeout=1800s @ 2026-07-01 12:51:12 =====
```

TaoTao-real:
Camodel/msprof measurement update for the #870 identity-TMOV pattern:

- Ran a focused AIV/vector microbenchmark on the A3 remote host using `task-submit + msprof op simulator`.
- Baseline reproduces 320 iterations of the redundant pattern fixed here: `pipe_barrier(PIPE_V); TMOV(alias, acc)` where both tiles are bound to the same UB address (`0x1000`).
- Candidate removes that identity-TMOV pattern. Common `TLOAD/TADD/TSTORE` work and public synchronization are kept identical.
- Both runs passed correctness: `repeat=320 expected=2 first=2 max_abs_err=0`.

Results:

| Variant | core0.veccore0 duration | Total tick | redundant `BAR PIPE:VEC` | redundant `MOV_UB_TO_UB` |
|---|---:|---:|---:|---:|
| baseline | 3.56 us | 8123 | 320 | 320 |
| candidate | 0.62 us | 2672 | 0 | 0 |

Delta on this focused pattern: `-2.94us` (`-82.6%`) and `-5451` simulator ticks (`-67.1%`).

Artifacts are under:
`/data/pyptouser/zhongxuan/ptoas_board_runs/issue870_identity_tmov_camodel/`

Scope note: this is a pattern microbenchmark for the #870 dead self-copy, not a full Qwen decode end-to-end performance claim.

TaoTao-real:
Follow-up update after the camodel/msprof measurement comment:

- The actual camodel/msprof data remains in the PR body and previous comment: baseline `3.56us / 8123 ticks / 320 BAR PIPE:VEC / 320 MOV_UB_TO_UB`, candidate `0.62us / 2672 ticks / 0 / 0`, both correctness-passing with `max_abs_err=0`.
- I also pushed `227326ce` to keep the identity-TMOV cleanup conservative for A5 low-precision `TMOV` cases. The pass now skips operands/results whose element type is in `isPTOLowPrecisionType` (`hif8`/`fp4` family), so the existing low-precision EmitC tests should remain protected while #870's plain identity self-copy pattern is still removed.
- Local validation available here: license header check OK, `git diff --check` OK, and targeted object compile OK for `PTORemoveIdentityTMov.cpp.o`. Full local `ptoas` rebuild is still blocked by the pre-existing local LLVM21 API mismatch around `Type::isFloat8*`/`getStridesAndOffset`, so CI is the authoritative end-to-end lit rerun for the low-precision cases.

TaoTao-real:
Addressed the P1 cast-semantics review in `0d2124f0`.

What changed:

- The identity-TMOV address evaluator no longer treats integer casts as transparent.
- It now evaluates integer-like constants through cast semantics explicitly:
  - `arith.index_cast` / `arith.extsi` use signed extension/truncation semantics.
  - `arith.index_castui` / `arith.extui` use unsigned extension/truncation semantics.
  - `arith.trunci` applies real truncation and otherwise bails out conservatively.
- If the cast chain cannot be modeled exactly, the address is treated as unknown and the `TMOV` is preserved.

Regression coverage added in `issue870_identity_tmov_pruning.pto`:

- `non_identity_tmov_extui_negative_addr`: proves `extui(arith.constant -1 : i8) -> i64` is `255`, not equal to `arith.constant -1 : i64`, so `TMOV` is preserved.
- `non_identity_tmov_trunc_then_extsi_addr`: proves `trunci(255 : i64) -> i8` followed by `extsi -> i64` is `-1`, not `255`, so `TMOV` is preserved.

Local validation:

- `git diff --check` OK
- license header check OK
- targeted object compile OK: `ninja -C build-llvm21 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o`

Full local `ptoas`/lit is still blocked on this machine by the pre-existing LLVM21 API mismatch around `Type::isFloat8*` / `getStridesAndOffset`; CI is the authoritative end-to-end rerun for the new lit coverage.

TaoTao-real:
CI follow-up for `build-and-test`:

- The only failed test was `PTOIR :: pto/issue870_identity_tmov_pruning.pto`.
- Root cause was test fragility, not the cast evaluator implementation: the `extui` regression was checked in the `--mlir-print-ir-after=pto-remove-identity-tmov` dump, but that dump is not a stable place to require every preserved/no-change function to appear.
- Pushed `cfea6ebb` to keep the `extui(i8 -1) -> i64` regression in the final C++ check (`TMOV(` must still be emitted), while removing the unstable IR-dump label for that function.
- The `trunci -> extsi` regression remains checked in the IR dump and the final C++ output.

This still catches the bug under review: if the pass again treats `extui(i8 -1)` as equal to `i64 -1` and erases the `TMOV`, the final C++ `TMOV(` check fails.

TaoTao-real:
Follow-up for the repeated `issue870_identity_tmov_pruning.pto` CI failure:

- The failure was caused by relying on `--mlir-print-ir-after=pto-remove-identity-tmov` for multiple function-level checks. That pass dump is not stable for requiring every preserved/no-change function to appear in source order.
- Pushed `9874c2fe` to make the regression check the final generated C++ output instead.
- This still validates the intended behavior directly:
  - identity same-address TMOV cases must not emit `TMOV(` or the induced `pipe_barrier(PIPE_V)`.
  - different-address, dynamic-address, `extui(i8 -1) -> i64`, and `trunci(255) -> i8` + `extsi -> i64` cases must still emit `TMOV(`.

So the test remains sensitive to the P1 bug: if address casts are treated transparently again and the write-like `TMOV` is erased, the final-output `TMOV(` check fails.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
```diff
@@ -0,0 +1,362 @@
...
+  const BaseMemInfo *dstInfo = getSingleMemInfo(buffer2MemInfoMap, dst);
+  if (!hasExactSameAddressRange(srcInfo, dstInfo))
+    return false;
+
+  if (srcInfo->rootBuffer == dstInfo->rootBuffer)
+    return true;
+
+  auto srcRootAddr = tryGetConcreteRootAddress(srcInfo);
+  auto dstRootAddr = tryGetConcreteRootAddress(dstInfo);
+  return srcRootAddr && dstRootAddr && *srcRootAddr == *dstRootAddr;
```
zhangstevenunity: **Blocker -- this fallback can drop a real cross-pipe sync edge (under-synchronization).**

This branch erases an identity `tmov` when `src` and `dst` have *different* InsertSync root buffers but the same concrete physical address. That is exactly the configuration where InsertSync will **not** reconnect a producer of `src` to a consumer of `dst`, because for local (UB/L1) memory the aliasing test keys on **root-buffer identity, not concrete-address overlap**:

- `MemoryDependentAnalyzer::MemAlias` (`lib/PTO/Transforms/InsertSync/MemoryDependentAnalyzer.cpp:158-187`) returns `false` for two local-mem accesses whose `rootBuffer` differ *and* whose `GetRealRoot` differ, even when `baseAddresses` overlap exactly.
- `GetRealRoot` (`:48`) stops at `alloc_tile` / `pointer_cast`; it never unifies two distinct allocations.
- `UpdateAllocTileOpMemInfo` (`PTOIRTranslator.cpp:483-489`) sets `rootBuffer = res` (the alloc itself), so two distinct `alloc_tile`s at the same address *always* have different roots.

`TMovOp::getEffects` (`lib/PTO/IR/PTO.cpp:13034`) reads `src` and writes `dst`, so the `tmov` is the *only* op carrying accesses on **both** roots. With it present, InsertSync builds `producer -> tmov(read src-root)` and `tmov(write dst-root) -> consumer`. Delete it and the two roots no longer alias, so the `producer -> consumer` cross-pipe set/wait flag is never emitted.

Concrete reproducer -- the PR's own `identity_tmov_same_addr` shape, but with a consumer of `dst` on a different pipe:

```mlir
%acc   = pto.alloc_tile addr = %c0_i64 : !pto.tile_buf<loc=vec, ...>   // root A
%alias = pto.alloc_tile addr = %c0_i64 : !pto.tile_buf<loc=vec, ...>   // root B, same addr
pto.tadds ins(%acc, %one ...) outs(%acc ...)   // producer: writes root A on PIPE_V
pto.tmov  ins(%acc ...)       outs(%alias ...) // identity bridge (pruned here)
pto.tstore ... ins(%alias) ...                 // consumer: reads root B on PIPE_MTE3
```

- Before the pass: InsertSync emits `PIPE_V -> PIPE_MTE3` (tmov writes `alias`, tstore reads `alias`, same root) -- correct.
- After the pass: `tadds` writes root A, `tstore` reads root B; different roots, same address -> `MemAlias=false` -> **no flag -> RAW race on the UB address**.

This is live specifically at `--pto-level=level3` (the test level), where `PlanMemory` is skipped (`tools/ptoas/ptoas.cpp:1865`) so the two `alloc_tile`s keep distinct roots, and in any manually same-addressed pair, which is the #870 scenario (`alias`/`acc` bound to the same UB address). The existing tests pass only because their `dst` is dead.

Suggested fix -- gate deletion on the condition InsertSync can still model the dependency:

- only erase when `srcInfo->rootBuffer == dstInfo->rootBuffer` (drop the `tryGetConcreteRootAddress` cross-root fallback at 312-314), **or**
- only erase when `dst` has no uses other than this `tmov` (truly-dead self-copy, which is the actual #870 redundant-copy pattern).

Either keeps the no-op removal while making it impossible to delete a `tmov` that is the sole sync bridge between two distinct roots. Given the camodel/msprof harness is already wired up for this PR, the consumed-`dst` variant above would be a good thing to run through it.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+  auto pointerCast = value.getDefiningOp<PointerCastOp>();
+  if (!pointerCast || pointerCast.getAddrs().size() != 1)
+    return std::nullopt;
+
+  Value addr = pointerCast.getAddrs().front();
+  return AddressToken{addr, tryEvalI64Constant(addr)};
+}
+
+static bool sameAddressToken(const AddressToken &lhs, const AddressToken &rhs) {
```
zhangstevenunity: **Secondary (same hazard class, narrower).** `sameAddressToken` treats two tokens as identical when their *constants* are equal even if they are distinct SSA values from distinct roots. For `pointer_cast`, InsertSync's root is the address SSA operand (`PTOIRTranslator.cpp:523-529`, `rootBuffer = rootSrc`), so two `pointer_cast`s with equal-but-distinct-SSA constant addresses are distinct roots -- the same under-sync hole as the `alloc_tile` case at `:314`, just usually masked by CSE merging identical constants into one SSA value. The root-identity / dead-`dst` gate suggested at `:314` covers this too; prefer comparing the InsertSync root over the raw constant value.

zhangstevenunity (changes requested):
## REQUEST_CHANGES

Nice, conservative framing, and the integer cast-semantics modeling (signed/unsigned/trunc with explicit bailout) is solid. One soundness gap blocks it.

**The identity proof can delete a `tmov` that is the sole cross-pipe sync bridge between two distinct InsertSync roots that happen to share a physical address** -- i.e. under-synchronization on the default `--enable-insert-sync` path. Full mechanism, a reproducer, and two fix options are in the inline comment at `PTORemoveIdentityTMov.cpp:314`. The current regression tests do not catch it because every identity case has a dead `dst` (never read after the move), which is trivially safe; the dangerous case is a *consumed* `dst`.

Secondary (same hazard class, narrower) note inline at `:144`.

Minor: the pass is registered unconditionally (`tools/ptoas/ptoas.cpp:1873`, before the sync-mode switch), so it also alters GraphSyncSolver / BufidSync / barrier-all / no-sync codegen, not just InsertSync. The no-op removal is generally harmless there, but the cross-layer reach is worth a deliberate decision.

Note: the Codex bot review for this PR failed (`exit=1`, model at capacity) at the stale head `0d2124f0`, so there are no structured bot findings to reconcile against; this is an independent read of the current head `9874c2fe`.

TaoTao-real:
Follow-up for the same-address / different-root sync-bridge concern:

- I agree that simply changing the rule to `same root + same range` would be safe but would not cover the #870 pattern, where the redundant TMOV can come from different roots bound to the same concrete UB address.
- The pruning rule is now split:
  - same InsertSync root + same exact range: prune as a true identity TMOV;
  - different roots + same concrete address/range: prune only when the TMOV destination has no uses outside this TMOV and all TMOV results are unused.
- If the destination is live after the TMOV, the op is preserved, because it may be the only bridge that lets InsertSync see the dependency from the src root to the dst root.
- Added regression coverage for a live-dst same-address case that must keep `TMOV`, while the dead-dst #870-style case is still pruned.

Local checks on this follow-up:
- `git diff --check` OK
- `ninja -C build-issue870 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o` OK
- Full local `ptoas` rebuild is still blocked by the pre-existing LLVM API mismatch in unrelated IR files (`Type::isFloat8*`, `getStridesAndOffset`), so CI remains the full gate.

TaoTao-real:
Follow-up after the `build-and-test` failure:

- The failed check was in the new `live_dst_same_addr_keeps_tmov` test, not in the pass implementation.
- Root cause: the test reused the same `%c0_i64` address SSA for src and dst, so it did not actually model the reviewer scenario of different InsertSync roots with the same concrete address. In that same-token/same-root shape, removing the identity TMOV is expected and safe.
- Updated the test to use different SSA address expressions that both statically evaluate to 0 (`0:i64` vs `extsi(0:i32)->i64`). This now exercises the intended different-root / same-concrete-address / live-dst case.
- Also updated the pass description to document the live-dst bridge guard.

Local checks for this follow-up:
- `git diff --check` OK
- `ninja -C build-issue870 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o` OK

TaoTao-real:
Follow-up for the latest `build-and-test` failure:

- The previous live-dst regression still used address expressions that were folded/CSE'd into the same emitted address token, so the final code still represented the safe same-root shape and the identity TMOV was removed.
- Reworked `live_dst_same_addr_keeps_tmov` to use two distinct `pto.pointer_cast` roots that point at the same address, then consume `%dst` through a second TMOV.
- The final check now requires two `TMOV(` calls in that function. If the first same-address live bridge is incorrectly deleted, only the second TMOV remains and the test fails.

Local checks:
- `git diff --check` OK
- `ninja -C build-issue870 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o` OK

TaoTao-real:
Follow-up for the latest `build-and-test` failure:

- CI correctly exposed that the previous implementation still allowed a same-address TMOV to be pruned through the MemInfo path when the roots collapsed to the same InsertSync root token. That bypassed the live-destination bridge guard and left only the downstream TMOV in `live_dst_same_addr_keeps_tmov`.
- Pushed `1ed680d3` to remove the explicit-address shortcut for non-self copies and route all non-`src == dst` identity decisions through `BaseMemInfo` plus the dead-destination guard.
- The pruning rule is now: exact same-SSA self-copies may be removed directly; every other same-address/range TMOV is removed only when the destination/result is dead and there is no later same-address alias use that could rely on the TMOV as a sync bridge.
- I also updated the pass description to match that conservative rule.

Local validation on this follow-up:
- `git diff --check` OK
- `ninja -C build-issue870 lib/PTO/Transforms/CMakeFiles/obj.PTOTransforms.dir/PTORemoveIdentityTMov.cpp.o` OK
- Full local `ptoas` link is still blocked by the existing local LLVM header/API mismatch in unrelated IR files, so GitHub CI remains the end-to-end gate.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
TaoTao-real: Thanks, this was a real soundness gap. Addressed in `1ed680d3`.

The current rule no longer uses the cross-root concrete-address fallback as a standalone identity proof for non-self copies. Except for exact `src == dst` self-copies, every same-address/range `tmov` now goes through `BaseMemInfo` and is pruned only when:

- the TMOV destination has no uses outside this TMOV and all TMOV results are unused; and
- there is no later same-address alias use that could rely on this TMOV as the only InsertSync bridge.

So the consumed-`dst` case in this comment is preserved now. The regression `live_dst_same_addr_keeps_tmov` models two distinct `pointer_cast` roots at the same address, consumes `%dst` through a later TMOV, and requires both `TMOV(` calls to remain in final C++ output. The latest CI is green, including `build-and-test`.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
TaoTao-real: Also addressed by the same follow-up in `1ed680d3`.

`sameAddressToken` / the explicit-address shortcut has been removed for non-self copies, so equal constant address values can no longer bypass the InsertSync-root / dead-destination guard. Equal-but-distinct address SSA values now only help prove physical same-address/range through `BaseMemInfo`, and deletion still requires the destination/result to be dead plus no later same-address alias use.

The `live_dst_same_addr_keeps_tmov` regression specifically covers the `pointer_cast` variant with distinct roots pointing at the same concrete address and keeps the bridge TMOV when `%dst` is consumed.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
```diff
@@ -0,0 +1,134 @@
...
+  auto pointerCast = value.getDefiningOp<PointerCastOp>();
+  if (!pointerCast || pointerCast.getAddrs().size() != 1)
+    return std::nullopt;
+
+  Value addr = pointerCast.getAddrs().front();
+  llvm::APInt constant;
+  if (matchPattern(addr, m_ConstantInt(&constant)))
+    return AddressToken{addr, constant};
+  return AddressToken{addr, std::nullopt};
+}
```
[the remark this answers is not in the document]
TaoTao-real: This thread is now outdated after the later safety fixes.

`getExplicitAddressToken` has been removed from the non-self deletion path. To avoid bypassing InsertSync-root semantics, non-`src == dst` identity decisions now go through `BaseMemInfo` and the dead-destination / no-later-alias-use guard. Constant address casts are still modeled exactly by `tryEvalIntegerLikeConstant` / `tryEvalI64Constant`, but they no longer form an unsafe standalone deletion proof.

lib/PTO/Transforms/PTORemoveIdentityTMov.cpp
```diff
@@ -0,0 +1,134 @@
...
+  }
+  return true;
+}
+
+static bool isIdentityTMov(TMovOp op) {
+  if (!hasPlainTMovSemantics(op) || !hasCompatibleIdentityTypes(op))
+    return false;
+
+  if (op.getSrc() == op.getDst())
+    return true;
```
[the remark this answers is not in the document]
TaoTao-real: This suggestion is intentionally not applied as a direct deletion rule after the sync-bridge review.

The current direct fast path is limited to exact `src == dst` self-copies. Metadata/view-like aliases are handled through `BaseMemInfo` instead, because peeling them and deleting unconditionally can hide whether a TMOV is the only bridge between InsertSync roots. This keeps the cleanup conservative while still allowing proven dead same-address copies to be removed.

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`d81c72b22a88`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_141526_merged_pr876.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_141526_merged_pr876.tsv`

PR #879, opened 2026-06-29
fix error in tload template for cube kernel

Likai-19:
修改 `lib/TileOps/tload_template.py`，分两部分：

**1. 消歧逻辑改为 layout 优先**

原来通过 rank-5 shape 做消歧（`src.shape[4] == dst.valid_shape[1]`），改为通过源 view 的 `layout` 属性（`src.config.layout`）。constraint 函数检查 `layout` 是否匹配预期（ND2NZ 只接受 `ND`/`NZ` 源，DN2NZ 只接受 `DN`/`NZ` 源，DN2ZN 只接受 `DN`/`NZ` 源）。当 `layout` 未知时保守返回 `True`（不消歧，保持旧行为）。

**2. 新增 `tload_gm_to_mat_dn2zn` 模板**

约束条件：
- `src.layout` 为 `DN` 或 `NZ`
- `dst.blayout=ROW_MAJOR, dst.slayout=COL_MAJOR`

模板实现参照 `TLoad.hpp:472` 的 `TLoadCubeDN2ZN`：
```python
n_value = g4                       # gShape4（最内层维度）
d_value = m                        # validRow（tile 行数）
src_inner_stride = s4 * elem_bytes # GetByteSize(gStride4)
pto.mte_gm_l1_frac(..., pto.FractalMode.ND2NZ,  # 复用 ND2NZ 硬件指令
    shape=(n_value, d_value),
    src_layout=(src_inner_stride,),
    dst_group=(1, 1, k, 0),
    ctrl=(0, False))
```

### DIFF
--- a/lib/TileOps/tload_template.py
+++ b/lib/TileOps/tload_template.py
@@ -353,15 +353,17 @@ def _constraint_tload_mat_nd2nz(src, dst) -> bool:
     # COL_MAJOR + ROW_MAJOR corresponds to NZ format
     if b_layout_value not in {"col_major", "COL_MAJOR"} or s_layout_value not in {"row_major", "ROW_MAJOR"}:
         return False
-    # ND2NZ: source is in ND (row-major) format where the inner dimension (g4)
-    # corresponds to the tile column count. Disambiguates from DN format where
-    # g4 corresponds to the tile row count.
-    if hasattr(src, 'rank') and src.rank == 5:
-        dst_valid_cols = dst.valid_shape[1] if hasattr(dst, 'valid_shape') and dst.valid_shape is not None else None
-        if dst_valid_cols is not None and hasattr(src, 'shape') and src.shape is not None:
-            src_inner = src.shape[4] if len(src.shape) >= 5 else None
-            if src_inner is not None:
-                if not _known_eq(dst_valid_cols, src_inner):
+    # Disambiguate ND vs DN via the source view layout attribute
+    # (e.g. layout = #pto.layout<nd>).  When the layout is not annotated the
+    # constraint conservatively returns True to avoid false rejections.
+    # src.config is always a ViewConfig for view operands — never None.
+    if hasattr(src, 'config') and src.config is not None:
+        view_config = src.config
+        if hasattr(view_config, 'layout'):
+            layout = view_config.layout
+            if layout is not None:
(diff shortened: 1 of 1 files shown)


Likai-19:
Fixes #877

reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #879 fix error in tload template for cube kernel
- Author: `Likai-19`
- Base/Head: `main` / `bugfix_tload_cube_kernel`
- Head SHA: `c2fb532d78ed`
- Trigger: 检测到新的 open PR
- Generated At: `2026-06-29T14:05:01Z`
- Status: completed

### Summary

Layout-based TLOAD.MAT matching in PR #879 regresses existing DN2NZ selection and also breaks MAT-NZ template resolution when view layout inference is unavailable.

### Findings

1. **P1** Existing DN2NZ loads are rematched to ND2NZ under inferred layouts `lib/TileOps/tload_template.py`:`386`

`_constraint_tload_mat_dn2nz` now disambiguates only from `src.config.layout`. That breaks the repo's existing DN2NZ call pattern, which uses transposed 2-D views like `shape=[32,16], strides=[16,1]` without an explicit `layout=#pto.layout<dn>` (see `test/lit/pto/cube_tile_ops_positive.pto` and `test/tilelang_st/npu/a5/src/st/testcase/tload_mat/tload_mat.pto`). `InferPTOLayout` classifies that shape/stride pair as ND, so after this change `_constraint_tload_mat_dn2nz` rejects the case while `_constraint_tload_mat_nd2nz` accepts it. The expanded kernel will therefore emit ND2NZ for an existing DN2NZ input, changing source addressing from logical `src[d,n]` to `src[n,d]` and loading the wrong matrix contents.

2. **P2** MAT-NZ tload becomes ambiguous when layout inference is disabled `lib/TileOps/tload_template.py`:`357`

Both MAT-NZ constraints now intentionally return `True` when the source view has no `config.layout`. Because the ND2NZ and DN2NZ kernels still have identical priorities and dtype signatures, any build that uses the supported `--disable-infer-layout` path can no longer disambiguate `pto.tload` into MAT/NZ: `select_kernel()` will see both templates as highest-priority matches and fail. The previous shape-based checks did not depend on layout inference, so this is a real compatibility regression for users who compile without inferred layouts.

PR #887, opened 2026-07-01
fix(emitc): preserve private helper static linkage

### LINKED ISSUE #884: 0.47 regression: `func.func private` lowered without `static` (external linkage) → duplicate-symbol collisions, AICORE_EXCEPTION 507015
## Summary

Starting in **ptoas 0.47**, a `func.func private` (internal-linkage) function is
lowered to C++ **without the `static` qualifier**, giving it *external* linkage.
0.45 and 0.46 correctly emit it as `static`.

When two kernels that each carry a private helper of the **same name** (a common
pattern for the distributed comm helpers `CommRemoteOffset_i32` / `_bf16` /
`_i8` / `_f32`, which PyPTO emits into every comm kernel) are compiled to
separate `.o` and linked into one device program, the now-external symbols
**collide (ODR / duplicate definition)**. The linked kernel is corrupted and
faults at runtime with `ACL_ERROR_RT_AICORE_EXCEPTION (507015)`.

## Impact

PyPTO DeepSeek-V4 distributed MoE (`--ep 2`, two ranks) compiles and runs
cleanly with ptoas 0.45/0.46, but with 0.47 it crashes on device during stream
sync:

```
[ERROR] sync_run_streams: aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507015
[ERROR] AICore error 507015: bounded device drain failed: 507015
```

Both `combine_push` and `dispatch` (linked together) define
`func.func private @CommRemoteOffset_i32`. Under 0.47 each `.o` exports it as a
**global** symbol:

```
# nm combine_push.o   (ptoas 0.47)
0000000000000000 T _Z20CommRemoteOffset_i32PU3AS1ll      <- global, should be local
# nm dispatch.o       (ptoas 0.47)
000000000000002c T _Z20CommRemoteOffset_i32PU3AS1ll      <- duplicate global
```

Under 0.45/0.46 the same symbol is `static` (file-local), no collision.

## Minimal reproducer

`repro.pto`:

```mlir
module {
  func.func @k(%arg0: !pto.ptr<i64>, %arg1: index) attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
    %0 = func.call @my_helper(%arg1) : (index) -> index
    func.return
  }
  func.func private @my_helper(%x: index) -> index {
    %c = arith.constant 1 : index
    %r = arith.addi %x, %c : index
    return %r : index
  }
}
```

```
ptoas repro.pto -o repro.cpp --enable-insert-sync --pto-level=level3 --pto-arch a3
grep my_helper repro.cpp
```

| ptoas | output for the `private` helper |
| ----- | ------------------------------- |
| 0.45  | `static AICORE int64_t my_helper(int64_t v1) {`  ✅ |
| 0.46  | `static AICORE int64_t my_helper(int64_t v1) {`  ✅ |
| 0.47  | `AICORE int64_t my_helper(int64_t v1) {`  ❌ (missing `static`) |

## Expected vs actual

- **Expected:** `func.func private @foo` lowers to an internal-linkage C++
  definition (`static ...`), as in 0.45/0.46.
- **Actual (0.47):** the `private` symbol visibility is ignored and the function
  is emitted with external linkage, causing duplicate-symbol collisions when
  multiple kernels carry same-named private helpers.

## Notes

The faulty entry-point handling around the same area also changed in 0.47
(0.45 emitted `extern "C" __global__ AICORE void <kernel>(...)`, 0.47 emits
plain `AICORE void <kernel>(...)`), but PyPTO normalizes the entry qualifier, so
that part is harmless. The **dropped `static` on `private` functions** is the
regression that breaks linking. A fix that re-honors MLIR `sym_visibility =
"private"` (→ `static`) restores correct behavior.

Environment: aarch64, CANN 9.0.0, `--pto-arch a3` (Ascend 910B). The missing
`static` reproduces independently of `--pto-arch` and `--enable-insert-sync`.

Zhendong404:
## Summary
- restore `static AICORE` emission for `func.func private` helpers in EmitC lowering
- add a regression test for issue #884 covering private helper static linkage
- keep explicit `pto.visibility = "external"` handling for public helper C linkage unchanged

## Validation
- `build/tools/ptoas/ptoas test/lit/pto/emitc_private_helper_static_linkage.pto --enable-insert-sync --pto-level=level3 --pto-arch a3 | /home/zhangzhendong/ptoas-workspace/llvm-project/build-shared/bin/FileCheck test/lit/pto/emitc_private_helper_static_linkage.pto`
- `build/tools/ptoas/ptoas test/lit/pto/emitc_public_helper_c_linkage.pto | /home/zhangzhendong/ptoas-workspace/llvm-project/build-shared/bin/FileCheck test/lit/pto/emitc_public_helper_c_linkage.pto`

Closes #884

### DIFF
--- a/lib/PTO/Transforms/PTOToEmitC.cpp
+++ b/lib/PTO/Transforms/PTOToEmitC.cpp
@@ -3296,6 +3296,9 @@ struct FuncToEmitC : public OpConversionPattern<func::FuncOp> {
     if (pto::isPTOEntryFunction(op)) {
       emitcFunc.setSpecifiersAttr(
           rewriter.getStrArrayAttr({"extern \"C\"", "__global__ AICORE"}));
+    } else if (op.isPrivate()) {
+      emitcFunc.setSpecifiersAttr(
+          rewriter.getStrArrayAttr({"static", "AICORE"}));
     } else if (pto::hasExternalArtifactVisibility(op)) {
       emitcFunc.setSpecifiersAttr(
           rewriter.getStrArrayAttr({"extern \"C\"", "AICORE"}));
--- a/test/lit/pto/emitc_private_helper_static_linkage.pto
+++ b/test/lit/pto/emitc_private_helper_static_linkage.pto
@@ -1,0 +1,28 @@
+// Copyright (c) 2026 Huawei Technologies Co., Ltd.
+// This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+// CANN Open Software License Agreement Version 2.0 (the "License").
+// Please refer to the License for details. You may not use this file except in compliance with the License.
+// THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+// See LICENSE in the root of the software repository for the full text of the License.
+
+// RUN: ptoas %s --enable-insert-sync --pto-level=level3 --pto-arch a3 | FileCheck %s
+// Guards issue #884: MLIR private function visibility must lower to C++ static
+// linkage, otherwise same-named private helpers collide across object files.
+
+module {
+  func.func @k(%arg0: !pto.ptr<i64>, %arg1: index) attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
+    %0 = func.call @my_helper(%arg1) : (index) -> index
+    func.return
+  }
+
+  func.func private @my_helper(%x: index) -> index {
+    %c = arith.constant 1 : index
+    %r = arith.addi %x, %c : index
+    return %r : index
+  }
+}
+
+// CHECK-LABEL: static AICORE int64_t my_helper(
+// CHECK-LABEL: AICORE void k(
+// CHECK: my_helper(


lib/PTO/Transforms/PTOToEmitC.cpp
```diff
@@ -3296,6 +3296,9 @@ struct FuncToEmitC : public OpConversionPattern<func::FuncOp> {
     if (pto::isPTOEntryFunction(op)) {
       emitcFunc.setSpecifiersAttr(
           rewriter.getStrArrayAttr({"extern \"C\"", "__global__ AICORE"}));
+    } else if (op.isPrivate()) {
```
zhangstevenunity, line 3299: Non-blocking follow-up (pre-existing, out of scope for this PR). The LLVM backends handle linkage via `applyArtifactVisibilityLinkage` (VPTOLLVMEmitter.cpp / VPTOCANN900LLVMEmitter.cpp), which keys off `hasExternalArtifactVisibility` alone: every non-external function -- private AND public-without-`pto.visibility` -- becomes `InternalLinkage`. This patch only makes `private` static, so a `func.func public @helper` without `pto.visibility="external"` still falls through to the `else` and is emitted as plain `AICORE` (external, mangled). That is exactly the `helper_internal` case that emitc_visibility_controls_c_linkage.pto documents as intended-internal, so EmitC still emits an externally-linked symbol for it and stays open to the same ODR-collision class as #884 (for public-non-external helpers rather than private ones).

Consider keying this branch off `!hasExternalArtifactVisibility(op)` (i.e. make the `else` branch `static AICORE`) so EmitC matches the LLVM path and that documented contract. That broadens behavior beyond the reported regression, so leaving it as a follow-up is reasonable. The `isPrivate()`-before-`hasExternalArtifactVisibility` ordering here is fine and consistent with driver.cpp:428.

zhangstevenunity (approved):
## Review: correct, minimal regression fix -- LGTM

Verified this restores the exact pre-4a794d5d7 behavior (`func.func private` -> `static AICORE`), which is the regression #884 reports: private comm helpers like `CommRemoteOffset_i32` gained external linkage and collided across `.o` files (ODR -> AICORE_EXCEPTION 507015).

What I checked:
- The added branch sits after `isPTOEntryFunction` and before `hasExternalArtifactVisibility`. Ordering is safe because no producer emits `private` + `pto.visibility="external"`: driver.cpp:428 strips external on private (`visibility != "private"` guard), and PyPTO session.py:739-740 sets external only alongside `public`.
- The new test is a genuine guard, not tautological: without the patch `my_helper` emits `AICORE int64_t my_helper(` and the `static AICORE` CHECK-LABEL fails; with it, it passes. The callee-before-caller emission order the CHECK-LABEL sequence relies on matches the existing emitc_public_helper_c_linkage.pto convention.
- No existing lit test breaks: pto_entry_multifunc.pto line 20 `AICORE int32_t helper(` (a private helper) still substring-matches the new `static AICORE int32_t helper(`, and there is no `CHECK-NOT: static` or line-anchored specifier check anywhere in test/lit.
- License header present on the new file; this is a pure lowering restore, so no ODS/verifier/python/docs cross-layer sync is required.

One non-blocking consistency note left inline.

reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #887 fix(emitc): preserve private helper static linkage
- Author: `Zhendong404`
- Base/Head: `main` / `fix/issue-884-private-helper-static`
- Head SHA: `9c9a10ab195a`
- Trigger: 检测到新的 open PR
- Generated At: `2026-07-01T06:03:37Z`
- Status: completed

### Summary

PR #887 regresses EmitC linkage semantics by forcing `static` on private helpers even when the IR explicitly marks them as externally visible.

### Findings

1. **P2** \`private\` now overrides explicit external artifact visibility in EmitC `lib/PTO/Transforms/PTOToEmitC.cpp`:`3299`

The new `op.isPrivate()` branch runs before `pto::hasExternalArtifactVisibility(op)`. PTO’s contract says non-entry artifact visibility is controlled by `pto.visibility = "external"`, and the LLVM emitters still honor that attribute regardless of MLIR symbol visibility. After this change, `func.func private @helper() attributes {pto.visibility = "external"}` lowers to `static AICORE ...`, so the helper stops being exported from the generated object even though the IR explicitly requested external artifact visibility. That is a backend contract mismatch and can break callers that rely on `setExternalArtifactVisibility` or hand-authored IR. The `static` path should be gated by `!pto::hasExternalArtifactVisibility(op)` or moved after the external-visibility check.

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`a45233eef74a`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_151816_merged_pr887.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260701_151816_merged_pr887.tsv`

PR #908, opened 2026-07-07
Feature/a2a3 llvm

castigli:
# A2/A3 VPTO Elementwise PR Summary

## Summary

This branch adds the A2/A3 VPTO UB lowering path for tile elementwise operations and validates it end-to-end on A3 hardware. The implementation routes A2/A3 through the planned-memory UB pipeline (from A5 pipeline), lowers supported PTO tile ops to VPTO UB ops, and emits LLVM intrinsics.

## What Changed

- Added the A2/A3 VPTO UB lowering pipeline:
  `PTOViewToMemref -> PTOPlanMemory -> PTOResolveReservedBuffers -> PTOMaterializeTileHandles -> LowerPTOToUBufOps`.
- Implemented A2/A3 UB lowering for binary elementwise ops:
  `tadd`, `tsub`, `tmul`, `tdiv`, `tmax`, `tmin`, `tand`, `tor`, `txor`, `tshls`, and `tshrs`.
- Implemented unary elementwise ops:
  `tabs`, `trelu`, `tneg`, `texp`, `tlog`, `tsqrt`, `trsqrt`, and `trecip`.
- Implemented scalar-tile elementwise ops:
  `tadds`, `tmuls`, `tmaxs`, and `tmins`.
- Added fused elementwise support for `taddrelu`, including PTO IR, UB IR, lowering, LLVM intrinsic emission, PTODSL wrappers, and tests.
- Added UB/LLVM support for new VPTO UB ops and intrinsics, including `vaddrelu`, `vdup`, and `vln`.
- Integrated planned UB address allocation so UB lowering consumes planner-assigned `alloc_tile addr` values and models scratch usage for ops such as `txor`.
- Extended dispatch coverage for small, normal, count, row-repeat, tail, chunked, and multi-row shapes.
- Extended the PTODSL elementwise surface through helpers and `pto.tile.*` APIs.
- Added lit and hardware e2e coverage for binary, unary, scalar, bitwise, shift, fused, log, and reciprocal cases.

## Validation

Verified after merging current `main` with the LLVM 21 docker image:

- Build: passed
- VPTO UB lit: `60/60`
- Runtime toolchain pytest: `4/4`
- A3 unary hardware e2e: `80/80`
- A3 binary hardware e2e: `192/192`
- A3 scalar hardware e2e: `120/120`

Total A2/A3 shared VPTO elementwise hardware coverage validated on A3: `392` tests.


reedhecre:
## Codex Review

_该评论由 review 机器人自动更新。_

- PR: #908 Feature/a2a3 llvm
- Author: `castigli`
- Base/Head: `main` / `feature/a2a3-llvm`
- Head SHA: `2ca10958ae5f`
- Trigger: PR 有新提交
- Generated At: `2026-07-14T11:15:50Z`
- Previous Head SHA: `7f54d1f1ed56`
- Status: failed at `codex-review` (exit=1)

### Summary

Review failed at stage `codex-review`: exit=1

### Findings

### Log Tail

```text
 .../vpto/ub/ub_to_llvm/elementwise/vadd_dtypes.pto |   87 +
 .../vpto/ub/ub_to_llvm/elementwise/vadd_packed.pto |   49 +
 .../vpto/ub/ub_to_llvm/elementwise/vaddrelu.pto    |   57 +
 test/lit/vpto/ub/ub_to_llvm/elementwise/vdiv.pto   |   30 +
 test/lit/vpto/ub/ub_to_llvm/elementwise/vdup.pto   |   27 +
 test/lit/vpto/ub/ub_to_llvm/elementwise/vln.pto    |   26 +
 .../vpto/ub/ub_to_llvm/elementwise/vmax_vmin.pto   |   64 +
 test/lit/vpto/ub/ub_to_llvm/elementwise/vmul.pto   |   30 +
 test/lit/vpto/ub/ub_to_llvm/elementwise/vsub.pto   |   30 +
 .../lit/vpto/ub/ub_to_llvm/mask/set_mask_count.pto |   47 +
 test/lit/vpto/ub/ub_to_llvm/mask/set_mask_full.pto |   26 +
 test/lit/vpto/ub/ub_to_llvm/mask/set_mask_tail.pto |   25 +
 .../lit/vpto/ub/ub_to_llvm/transfer/copy_gm_ub.pto |   26 +
 .../vpto/ub/ub_to_llvm/transfer/copy_gm_ub_pad.pto |   25 +
 .../lit/vpto/ub/ub_to_llvm/transfer/copy_ub_gm.pto |   25 +
 test/lit/vpto/ub/ub_to_llvm/transfer/mte_gm_ub.pto |   25 +
 .../vpto/ub/ub_to_llvm/transfer/mte_gm_ub_pad.pto  |   29 +
 tools/ptoas/ObjectEmission.cpp                     |   30 +-
 tools/ptoas/driver.cpp                             |    6 +-
 tools/ptoas/ptoas.cpp                              |   92 +-
 104 files changed, 15327 insertions(+), 132 deletions(-)
===== END STAGE clone rc=0 @ 2026-07-14 19:15:40 =====

===== STAGE codex-review @ 2026-07-14 19:15:40 =====
set -euo pipefail
cd '/tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/repo'
'codex' exec -C '/tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/repo' -s read-only -c 'model_provider="codereview"' -c 'model="gpt-5.4"' -c 'model_reasoning_effort="xhigh"' --output-schema '/tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/review_schema.json' -o '/tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/codex_last_message.json' --color never - < '/tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/review_prompt.txt'
[monitor] stage timeout: 1800s
OpenAI Codex v0.115.0 (research preview)
--------
workdir: /tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/repo
model: gpt-5.4
provider: codereview
approval: never
sandbox: read-only
reasoning effort: xhigh
reasoning summaries: none
session id: 019f6056-ff5d-7433-b822-27cea04a9f17
--------
user
你现在在审查 GitHub PR。

仓库：hw-native-sys/PTOAS
PR：#908 Feature/a2a3 llvm
作者：castigli
base branch：origin/main
head branch：HEAD（当前已 checkout 到 PR head）

要求：
1. 只审查这个 PR 相对 origin/main 的改动，必要时可以看上下文文件。
2. 重点找真实的 correctness / regression / contract mismatch / CI / runtime / compatibility 问题。
3. 不要提纯风格建议，不要提低价值猜测。
4. 严格按优先级输出：
   - P1：高概率会导致错误结果、编译/运行失败、严重回归、发布阻断
   - P2：重要缺陷、行为回归、遗漏校验/测试、较大兼容性问题
   - P3：次要但明确可改的问题
5. 如果没有问题，summary 直接写：未检查到 PR #908 存在问题，并返回 findings=[]。
6. 如果有问题，summary 简洁概括，findings 里每条都要给出：
   - severity
   - title
   - body（说明为什么是问题，尽量具体）
   - file（尽量给相对路径）
   - line（能确定就填整数，否则 null）

建议先查看：
- git status --short
- git diff --stat origin/main...HEAD
- git diff --unified=80 origin/main...HEAD

最终输出必须严格匹配 JSON schema。

mcp startup: no servers
Reconnecting... 1/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c6eafebf04d-LAS, request id: c06f2d39-82c0-4a24-994e-659c73455fde)
Reconnecting... 2/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c71ce1d2acc-LAX, request id: 8d735904-af43-4c78-91c6-2e9f5f571ba4)
Reconnecting... 3/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c763b05cb96-LAX, request id: 9875af03-fd86-4c28-84f8-5616ac06e989)
Reconnecting... 4/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c7cc8a6b82d-LAS, request id: d4417066-987c-4621-abdc-3397bd18c85a)
Reconnecting... 5/5 (unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c887d4db171-LAS, request id: c4b7cf46-8fcb-4287-93f4-21a6d4e07c0a)
ERROR: unexpected status 403 Forbidden: {"code":"INSUFFICIENT_BALANCE","message":"Insufficient account balance"}, url: https://codex.0u0o.com/responses, cf-ray: a1b01c9f3f1dcba7-LAX, request id: c246dfe4-c841-440b-aa26-92af63ecadd9
Warning: no last agent message; wrote empty content to /tmp/ptoas-pr-review-monitor/runs/20260714_191529_pr908/codex_last_message.json
===== END STAGE codex-review rc=1 @ 2026-07-14 19:15:50 =====
```

ptodsl/examples/bop_launch_a3.py
```diff
@@ -0,0 +1,253 @@
...
+import time
+from pathlib import Path
+import sys
+
+import numpy as np
+
+if __package__ in {None, ""}:
+    here = Path(__file__).resolve()
+    for candidate in here.parents:
+        if (candidate / "ptodsl" / "__init__.py").exists():
```
Zhendong404: In latest main branch, you can install ptodsl package to the python env with `pip install . --no-build-isolation`. No need to manually search ptodsl path.

ptodsl/examples/bop_launch_a3.py
```diff
@@ -0,0 +1,253 @@
...
+            sys.path.insert(0, str(candidate))
+            break
+    else:
```
Zhendong404: We already have some ST cases in `test/tilelib-st`. Maybe you can reuse them by just changing the target from `a5` to `a3`.

ptodsl/ptodsl/_ops.py
```diff
@@ -2586,6 +2586,15 @@ def tadd(src0, src1, dst):
     )
 
 
+def taddrelu(src0, src1, dst):
```
Zhendong404, line 2962: Is this a A2A3-only op? Should restrict its usage inside A2A3 kernels.

ptodsl/tests/e2e/common.py
```diff
@@ -0,0 +1,486 @@
+# Copyright (c) 2026 Huawei Technologies Co., Ltd.
```
Zhendong404, line 1: The naming of this test suite is not clear. I suggest reuse the TileOp ST cases we already have in tilelib-st, which will be completed soon.

For vpto level test, you should place them under the `test/vpto` directory.

ptodsl/ptodsl/_ops.py
Zhendong404: Please also add documents for this new interface in the `ptodsl/docs/user_guide` directory.

Zhendong404 (changes requested):

docker/Dockerfile.dev
```diff
@@ -0,0 +1,84 @@
...
+# =========================================================================
+# 4. Clone pto-isa (headers for bisheng compilation of generated kernels)
+# =========================================================================
+RUN git clone "${PTO_ISA_REPO}" "${PTO_LIB_PATH}" \
+    && cd "${PTO_LIB_PATH}" && git checkout "${PTO_ISA_COMMIT}"
+
+# =========================================================================
+# 5. Torch + torch-npu (for on-device kernel launch)
+# =========================================================================
+RUN pip install --no-cache-dir torch==2.9.0 --index-url https://download.pytorch.org/whl/cpu \
```
zouzias, line 70: Better use `torch==2.9.0.post2` or `2.10.0` here for CANN-9.0.0, see compatibility matrix here: https://gitcode.com/Ascend/pytorch/blob/master/README.zh.md#ascend-extension-for-pytorch%E7%89%88%E6%9C%AC%E9%85%8D%E5%A5%97%E8%A1%A8

castigli: Good catch. The CANN 9.0.0 matrix pairs Ascend Extension for PyTorch 2.9.0.post2 with upstream PyTorch 2.9.0, so I kept `torch==2.9.0` and updated `torch-npu` to `2.9.0.post2`. I also updated the non-dev Dockerfile for consistency.

ptodsl/ptodsl/_ops.py
castigli: Addressed. I documented `pto.tile.addrelu` in `ptodsl/docs/user_guide/08-compute-operations.md`, including formula, supported dtypes, and the A2/A3-only restriction.

ptodsl/tests/e2e/common.py
castigli: I can move these tests under `test/tilelib-st` if that is the preferred repository layout, but I’m worried we would lose clarity and potentially coverage.

The current pytest suite is a comprehensive PTODSL e2e sweep for the new A2/A3 VPTO elementwise path. It covers binary, unary, scalar, bitwise, shift, fused, log, and reciprocal ops across multiple dtypes and dispatch shapes.

My concern with `test/tilelib-st` is that the folder name is not very descriptive for this purpose: these are specifically PTODSL end-to-end tests for the A2/A3 VPTO lowering pipeline. Moving them there may make that less obvious.

If you still prefer the ST layout, I can migrate them, but I’d like to preserve the current breadth of coverage.

ptodsl/ptodsl/_ops.py
castigli: Yes, this is A2/A3-only. I added a PTODSL tracing-time target guard, an IR verifier rejection for A5, and regression coverage for both PTODSL diagnostics and hand-authored PTO IR.

ptodsl/examples/bop_launch_a3.py
castigli: Addressed. I removed the manual parent-directory `sys.path` probing from the example; it now assumes PTODSL is installed in the Python environment.

castigli:
Retriggering CI for latest head c80521f13.

learning-chip:
How many "non-elementwise" vector ops are left? (gather/scatter/reduce/...) Is it too big to also include them into this PR?

castigli:
> How many "non-elementwise" vector ops are left? (gather/scatter/reduce/...) Is it too big to also include them into this PR?

As you said, only the gather/scatter/reduce are missing. Not a large addition, I  will add them in this PR.

tools/ptoas/ptoas.cpp
```diff
@@ -2581,11 +2641,14 @@ static void lowerPTOToVPTOBackend(PassManager &pm, ModuleOp module, int argc,
...
 static pto::VPTOEmissionOptions
-buildVPTOEmissionOptions(const pto::CANNVersion &cannVersion) {
+buildVPTOEmissionOptions(const pto::CANNVersion &cannVersion,
+                         llvm::StringRef targetArch) {
   pto::VPTOEmissionOptions options;
   options.dumpVPTOIR = false;
   options.targetTriple = "hiipu64-hisilicon-cce";
   options.cannVersion = cannVersion;
+  std::string arch = normalizeArch(targetArch);
+  options.march = isA2A3Arch(arch) ? "dav-c220-vec" : "dav-c310-vec";
```
zhangstevenunity: **P2 (A5 VPTO regression, outside the PR's A2/A3 scope).** `buildVPTOEmissionOptions` now unconditionally sets `options.march`, so for A5 it becomes the non-empty string `"dav-c310-vec"`. On `main` this field was left empty.

Downstream, `makeDeviceEmissionOptions` (VPTOLLVMEmitter.cpp:11395) branches on `options.march.empty()`. With a non-empty march the **Cube** kernel kind now takes the `else` branch and gets `aicoreArch = defaultTargetCPU = march = "dav-c310-vec"` (vector); only `defaultTargetFeatures` resolves to the c310-cube string (line 11414-11415). `ObjectEmission.cpp::resolveTargetCPU` then reads that `"dav-c310-vec"` target-cpu attr and passes `-mcpu=dav-c310-vec` to bisheng for the cube object.

Net effect: `ptoas --pto-arch=a5 --pto-backend=vpto <kernel with a cube section>` emits the cube (AIC) device object targeting the vector (AIV) sub-core -- silent wrong-target codegen. On `main` the empty-march path correctly produced `dav-c310-cube`. A5 vector modules are unaffected. Not covered by the new tests (all a2/a3 vector).

Suggest only seeding march for a2/a3 and leaving it empty for a5 so the per-kind defaulting keeps working:

```cpp
if (isA2A3Arch(arch))
  options.march = "dav-c220-vec";
// a5: leave march empty; makeDeviceEmissionOptions picks vec/cube per kind
```

lib/PTO/Transforms/VPTOLLVMEmitterDispatcher.cpp
```diff
@@ -12,7 +12,7 @@
 namespace mlir::pto {
 
 static bool usesCANN900Lowering(const CANNVersion &cannVersion) {
-  return cannVersion >= CANNVersion::release(9, 0, 0);
+  return false; // Force Beta1 path until CANN900 emitter is stable for A3
```
zhangstevenunity: **P2 (behavior change beyond A2/A3 scope).** This flips `usesCANN900Lowering` from `cannVersion >= 9.0.0` to an unconditional `return false`, so *every* VPTO device emission -- including A5 on CANN 9.0.0+ -- is now routed through the Beta1 emitter instead of the CANN900 emitter, and `lowerVPTOModuleToLLVMModulesCANN900` becomes dead code. The comment scopes the intent to A3, but the gate is global (no arch/version check).

If A5 (or other) VPTO targets on CANN9 previously went through the CANN900 path and the two emitters are not at parity, this is a silent codegen change for them, buried in an A2/A3 PR. Consider gating on `isA2A3` (or on arch) rather than a blanket `return false`, or explicitly confirm A5 VPTO board/lit coverage still passes on CANN9 with Beta1.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -0,0 +1,1797 @@
...
+                       "PTOMaterializeTileHandles before LowerPTOToUBufOps");
+          signalPassFailure();
+          return;
+        }
+        builder.setInsertionPoint(op);
+        auto ptrTy = pto::PtrType::get(
+            ctx, tbTy.getElementType(),
+            pto::AddressSpaceAttr::get(ctx, pto::AddressSpace::VEC));
+        auto pc = builder.create<pto::CastPtrOp>(op.getLoc(), ptrTy, addr);
+        tileShapes[pc.getResult()] = SmallVector<int64_t, 2>(shape);
```
zhangstevenunity: **P2 (from code inspection -- the valid-shape dispatch is effectively dead).** This rewrite stores only the **padded** `shape` into `tileShapes`, and the `pto::PtrType` branch of `extractTileShapeInfoFromValue` (line 156-161) leaves `validShape` empty. After this rewrite every compute-op operand is a `PtrType`, so `extractTileShapeInfo` computes `vCols == cols` and `vRows == rows` for all alloc_tile-derived tiles -- which is every lit test and the common kernel case.

Consequences:

1. In `dispatch` (line 1437) the `vCols == cols` guard is then always true, so the entire non-continuous subtree -- `modeColVLAlign`, `modeCount2L`, `modeRowRpt`, `rowRptFast`/`rowRptChunked`, `headRows`/`tailRows`, and the `tailStride*` helpers -- is unreachable. The lit tests named after those modes (`tadd_col_vl_align`, `tadd_row_rpt[_tail]`, `tadd_chunked[_tail]`, and `tadd_count`) actually lower through `modeNorm1L`/`modeCount1L`/`modeSmall`; they pass only because the CHECKs are structural (no test pins the emitted `set_mask` count for a `v_col < cols` shape). So the valid-shape paths have no real coverage.

2. Every op is lowered against the padded width, so it reads/writes the padding columns (uninitialized UB) rather than just `[0, v_col)`. Valid-region results stay correct only because tiles are allocated at padded size (PlanMemory uses `getShape()`), but it processes up to `cols/v_col` x more data and touches uninitialized memory (e.g. `vln`/`vrsqrt`/`vdiv` over padding).

Separately, the unary/scalar/dup paths (`dispatchUnary`/`dispatchShift`/`dispatchDup`, ~line 933-1079) iterate flat over `vRows*vCols` with no per-row `cols` stride, unlike the binary `dispatch`. That is only safe because validShape is currently dropped (flat == full contiguous buffer). If a valid-shape-preserving `TileBufType` tile (e.g. a non-alloc block-arg) with `v_col < cols` and `v_row > 1` ever reaches a unary/scalar op, those paths would skip rows / process padding as valid (wrong result), because they lack the guard the binary path has.

Suggestion: capture the tile's valid shape into `tileShapes` (or key the dispatch off the `alloc_tile` `valid_row`/`valid_col`), and add a lit test that pins the `set_mask` count for a padded multi-row shape so these modes get real coverage.

ptodsl/ptodsl/_runtime/toolchain.py
```diff
@@ -155,10 +170,15 @@ def runtime_library_flags(*, sim_mode: bool = False) -> list[str]:
     return flags
 
 
-def aicore_arch_for_kernel_kind(kernel_kind: str) -> str:
+def aicore_arch_for_kernel_kind(kernel_kind: str, target_arch: str = "a5") -> str:
```
zhangstevenunity: **P3 (cross-layer completeness).** This adds the `target_arch` parameter (and `test_runtime_toolchain.py` calls it directly), but the only production caller -- `_kernel_compile_flags(kernel_kind)` in `native_build.py:126` -- calls `aicore_arch_for_kernel_kind(kernel_kind)` *without* `target_arch`, so it always defaults to `"a5"` and emits `--cce-aicore-arch=dav-c310-vec` for the `launch.cpp` wrapper. `build_native_library` threads `module_spec.target_arch` into the kernel-object compile (`_run_ptoas`) but not into `_compile_launch_cpp`.

So on A2/A3 the kernel object is built for `dav-c220` while the launch stub is built for `dav-c310`. The new unit test passes but never exercises the real call path, giving false confidence. (The A3 e2e apparently still links/runs, so the wrapper's arch may not be load-bearing, but the two objects should agree.) Thread `module_spec.target_arch` through `_compile_launch_cpp` -> `_kernel_compile_flags` -> `aicore_arch_for_kernel_kind`.

zhangstevenunity (commented):
## Review: A2/A3 VPTO elementwise

Deep pass over the lowering (`LowerPTOToUBufOps`), the LLVM intrinsic emitter, the CLI/pipeline wiring, the ODS/verifiers/effects, and the PTODSL surface. Overall the A2/A3 elementwise path is well-built: the intrinsic emission is internally consistent with the design docs, the new ops (`vaddrelu`/`vdup`/`vln`) and the `txor` tmp-scratch effects modeling are correct, and most of the earlier Gemini review was against a stale commit and is addressed at HEAD.

The findings that stand out are all cases where this A2/A3 PR reaches into the shared **A5 / VPTO device-emission** path:

- **P2** `buildVPTOEmissionOptions` now mis-targets A5 VPTO **cube** kernels (`dav-c310-vec` instead of `dav-c310-cube`). Inline on `ptoas.cpp:2651`.
- **P2** `usesCANN900Lowering` flipped to unconditional `false` disables the CANN900 emitter for **all** targets/CANN versions, not just A3. Inline on `VPTOLLVMEmitterDispatcher.cpp:15`.
- **P2** The lowering silently drops each tile's valid-shape, so the valid-shape-aware dispatch modes (and the lit tests named after them) are effectively dead, and every op processes padding. Inline on `LowerPTOToUBufOps.cpp:252`.
- **P3** The PTODSL native launch-wrapper compile doesn't receive `target_arch`. Inline on `toolchain.py:173`.

### Verified clean (not bugs at HEAD)

- `txor`'s `tmp` is a required operand and WRITE-modeled in `getEffects`; PlanMemory gives it a distinct buffer and `LowerPTOToUBufOps` decomposes `txor` atomically -> no aliasing/missing-sync bug (the recurring scratch-effect bug class is not present here).
- `taddrelu` A2/A3-only: `TAddReluOp::verify` + `dispatchVerifierByArch` reject a5, and the dtype gate `i16/f16/f32` matches the available `VADDRELU` intrinsics; PTODSL has a trace-time guard plus negative tests.
- `vaddrelu`/`vdup`/`vln` intrinsic names, operand order, and config-word layouts match the design doc and the lit tests; the CBuf storage-bit-width fix (2ca10958) is sound.
- Gemini's `== "dav-c220-vec"` cube-exclusion flag is a false alarm: the elementwise op-pattern/legality sites (10994/11166) are vector-only by design and `march` is never `dav-c220-cube` on this path. Gemini's leftover-lambda compile error and `VERITY` typo were on a stale commit and are not present at HEAD.

### Lower-priority notes

- UB op verifiers only check the UB memory-role, not that `dst`/`src` share element type/shape; a hand-written `.pto` with mismatched pointer element types picks the intrinsic from `src0` and miscompiles with no diagnostic (`VPTOLLVMEmitter.cpp:4744`). Compiler-generated IR is always consistent, so this is a robustness gap.
- `set_mask` hardware state is written by `UBSetMask*` but not read by the compute ops, so ordering is positional only; safe today (no scheduler on this IR post-lowering) but latent if one is added.
- `vdup` rejects bf16 (clean compile error, not a miscompile); scalar-tile bf16 would pass an i64 bit-pattern to `VMULS.bf16` (out of scope per doc, unguarded).
- `_ptoas_build_backend`: `build_editable(skip_install=True)` no longer validates the runtime payload landed where the `.pth` points; `_find_llvm_dir` auto-probe may select an Ascend-shipped LLVM over the vpto-patched one on env-unset machines.

### CI

The Codex review bot failed on `INSUFFICIENT_BALANCE` (not a code signal). No CI checks are reported on the branch, and the PR is currently `CONFLICTING` with `main`.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
zhangstevenunity: Only the physical `shape` is stored before `replaceAllUsesWith` converts every tile operand to `pto.ptr`. From then on `extractTileShapeInfoFromValue` takes the pointer path, has no `validShape`, and falls back to `vRows=rows`/`vCols=cols`. I reproduced this with the existing 4x96 f32, v_col=32 test: lowering emits six 64-element iterations (all 384 physical elements) instead of operating on 4x32 valid elements with a 96-element row stride. Please preserve both physical and valid shapes in the pointer metadata and make the test assert the iteration count and offsets.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -0,0 +1,1797 @@
...
+
+  template <typename UBop>
+  void modeCount1L(Location loc, OpBuilder &b, Value dst, Value s0, Value s1,
+                   pto::PtrType ptrTy, const TileShapeInfo &info) {
+    int64_t epr = info.elementsPerRepeat;
+    int64_t totalV = info.vRows * info.vCols;
+    int64_t totalRpts = (totalV + epr - 1) / epr;
+    b.create<pto::UBSetMaskCountOp>(loc);
+    b.create<pto::UBSetMaskOp>(loc, i64c(totalV, loc, b), i64c0(loc, b));
+    emitUBBinOp<UBop>(loc, b, dst, s0, s1, i64c(totalRpts, loc, b), i64c8(loc, b));
```
zhangstevenunity: `dispatch` calls this helper only when `totalRpts > 255`, yet the helper passes that value straight to the UB op. The 16x1024 f32 test lowers to `repeat = 256`; `LowerUBufBinary` masks repeat with `0xff`, so it is encoded as zero. The new design document also says C220 count mode does not support `repeat > 1`. Please split this into repeat-1 chunks or a loop and update the current `CHECK-NOT: scf.for` expectation.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -0,0 +1,1797 @@
...
+      return failure();
+    ArrayRef<int64_t> shape = memTy.getShape();
+    if (shape.size() < 2)
+      return failure();
+
+    OpBuilder b(ctx);
+    b.setInsertionPointAfterValue(view);
+    DmaViewInfo info;
+    auto ptrTy = pto::PtrType::get(ctx, memTy.getElementType(), msAttr);
+    Value root = traceRootMemRef(view);
```
zhangstevenunity: At this point `PTOViewToMemref` has already converted `partition_view` to `memref.subview`. Tracing to the root here discards the subview offset, while the following code rebuilds row-major strides and fills all offsets with zero, also discarding non-contiguous layouts. With a 17x32 view partitioned at [1, 0], lowering computes a zero byte offset instead of 128 bytes. Please extract strided metadata and accumulate the real view offsets and strides rather than reconstructing them.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -0,0 +1,1797 @@
...
+    fullMask(loc, b);
+  }
+
+  void setMask(Location loc, OpBuilder &b, unsigned n) {
+    auto [m0, m1] = computeContMaskValues(n);
+    b.create<pto::UBSetMaskOp>(loc, i64c(m0, loc, b), i64c(m1, loc, b));
+  }
+
+  void fullMask(Location loc, OpBuilder &b) {
+    b.create<pto::UBSetMaskOp>(loc, i64cM1(loc, b), i64c0(loc, b));
```
zhangstevenunity: `fullMask` emits `mask0=-1, mask1=0`, but this PR's own `UBSetMask` documentation says full-write mode is `-1, -1`; `mask1` controls lanes 64-127. Consequently, an f16 operation following a masked or count-mode operation can leave the upper half disabled. Please restore both mask words and add a chained masked-to-normal regression test.

ptodsl/ptodsl/_runtime/toolchain.py
zhangstevenunity: The new `target_arch` argument is never passed by native build: `_kernel_compile_flags` still calls `aicore_arch_for_kernel_kind(kernel_kind)`, and `_compile_launch_cpp` does not receive `module_spec.target_arch`. A2/A3 launcher code is therefore still compiled for `dav-c310-*` while PTOAS emits a C220 kernel. Please thread `target_arch` through the call chain and test the actual Bisheng command rather than only this helper.

zhangstevenunity (commented):
I found five correctness issues while reviewing the A2/A3 VPTO lowering and PTODSL integration. Inline details and reproductions are included below.

lib/PTO/Transforms/VPTOLLVMEmitter.cpp
```diff
@@ -10220,7 +11064,7 @@ static LogicalResult lowerVPTOTypes(ModuleOp module, llvm::raw_ostream &diagOS)
   LoweringState state;
 
   target.addLegalOp<ModuleOp>();
-  target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
+  target.addLegalOp<pto::AddPtrOp>();([&](func::FuncOp op) {
     return typeConverter.isSignatureLegal(op.getFunctionType()) &&
            typeConverter.isLegal(&op.getBody());
   });
```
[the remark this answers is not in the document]
castigli: Addressed. The legality callback is restored; the merged head builds and passes 919/919 lit tests.

lib/PTO/Transforms/VPTOLLVMEmitterDispatcher.cpp
castigli: Addressed in `419001469`. C220 remains on Beta1; non-C220 CANN 9 targets use the CANN900 emitter.

castigli:
Review update:

- Merged current `main` in `69be84f80` and resolved the upstream L2-cache API overlap.
- Local validation: 919/919 lit tests and 392/392 A3 VPTO hardware cases passed. Hardware testing was restricted with `ASCEND_RT_VISIBLE_DEVICES=2,3`.
- Gather is split into https://github.com/castigli/PTOAS/pull/2.
- Scatter and reduce will follow as separate PRs based on `feature/a2a3-llvm`.

The two test-layout discussions remain open intentionally. @Zhendong404, please re-review the refreshed head.

castigli:
@learning-chip I have a parallel PR adding gather ops (https://github.com/castigli/PTOAS/pull/2). I will do the same for scatter and reduce. However, given the cost keeping in sync with main, I would prefer to merge this PR first as it is and have 3 small follow ups for the missing vector ops. Is that OK?

learning-chip:
> I would prefer to merge this PR first as it is and have 3 small follow ups for the missing vector ops. Is that OK?

This is ok for me -- @zhangstevenunity ok to merge this?

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`c09a21312cb6`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260721_011321_merged_pr908.log`

### 失败用例
- `comm_p2p` (run, exit=1)

reedhecre:
## A3 板测失败详情：PR #908

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260721_011321_merged_pr908/npu_validation/CommSync/comm_p2p_binding_variants/main.cpp:122)
EZ9999[PID: 4083110] 2026-07-21-01:47:32.529.379 (EZ9999):  The error from device(chipId:0, dieId:1), serial number is 1766, there is an exception of aivec error, core id is 15, error code = 0x10, dump info: pc start: 0x124200000000, current: 0x124200000164, vec error info: 0x4000000b7, mte error info: 0x3c0300404d, ifu error info: 0x212c0bbbb5900, ccu error info: 0x21b59000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x300404d, fixp_error1 info: 0x3c, fsmId:0, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=1, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_binding_variants_kernel, fault kernel info ext=comm_p2p_binding_variants_kernel, program id=0, hash=16541545288638093324.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-21 01:47:33] ERROR: testcase failed (exit 1): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260721_011321_merged_pr908/npu_validation/CommSync/comm_p2p/main.cpp:122)
EZ9999[PID: 4093884] 2026-07-21-01:47:51.086.933 (EZ9999):  The error from device(chipId:0, dieId:1), serial number is 1767, there is an exception of aivec error, core id is 20, error code = 0x10, dump info: pc start: 0x124200000000, current: 0x124200000168, vec error info: 0xdf00005f59, mte error info: 0x6d0600009a, ifu error info: 0x212c0bbb3bdc0, ccu error info: 0x21b59000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600009a, fixp_error1 info: 0x6d, fsmId:0, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=1, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_kernel, fault kernel info ext=comm_p2p_kernel, program id=0, hash=6260346515940177617.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-21 01:47:53] ERROR: testcase failed (exit 1): comm_p2p
```

</details>

reedhecre:
## A5 板测成功
- 触发方式：`merged`
- 源码提交：`c09a21312cb6`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260721_183905_merged_pr908.log`
- 结果 TSV：`/root/ptoas-board-monitor-a5/logs/20260721_183905_merged_pr908.tsv`

PR #947, opened 2026-07-16
fix(ptodsl): correct AST rewrite loop scope

### LINKED ISSUE #933: [Bug] False-positive ast_rewrite runtime loop IV exposure for sibling range(...) loops reusing 'c'
### Component
Other (`ptodsl/ptodsl/_ast_rewrite.py` AST rewrite frontend)

### Description
`ast_rewrite=True` appears to raise a false positive for runtime `range(...)` loops when sibling runtime loops reuse the same induction variable name.

In the reproducer below, the first `for c in range(cols):` does not intentionally use `c` after the loop exits, but `probe.compile()` still fails with:

```text
ptodsl._ast_rewrite.PTODSLAstRewriteError: ast_rewrite=True runtime for-loops cannot expose the loop induction variable outside the loop yet; use explicit pto.for_(...) for 'c'
```

A practical workaround on my side is to rename the later sibling runtime loop variables (`c` -> `dpost_chunk`, `dcomb_chunk`, etc.). After renaming, the same kernel shape compiles.

So this looks like a frontend liveness / loop-IV analysis issue rather than a real user-side use-after-loop of the induction variable.

### Reproduction (minimal)
`repro.py`

```python
from ptodsl import pto

@pto.jit(target="a5")
def probe(rows: pto.i32, cols: pto.i32):
    mask1 = 1
    mask_all = 2
    for t in range(rows):
        for c in range(cols):
            col = c * 64
            mask = col
            d_heads = [mask + h for h in pto.static_range(4)]
            acc = None
            for h in pto.static_range(4):
                term = d_heads[h]
                acc = term if acc is None else acc + term
            sink = acc + col + mask

        for mo in pto.static_range(4):
            mix_acc = 0
            for c in range(cols):
                col = c * 64
                mask = col
                mix_acc = mix_acc + mask
            sum_v = mix_acc + mask_all
            out = sum_v + mask1

print(probe.compile().mlir_text())
```

Run:

```bash
python repro.py
```

Notes:
- If I rename the second loop variable from `c` to something else, the error goes away.
- I reduced this from a larger VMI kernel where several sibling `for c in range(chunks):` loops lived under the same outer `for t in range(num_tokens):` block.

### Expected behavior
The frontend should accept this pattern, or at least not report that `c` is exposed outside the loop when it is only reused as the induction variable of a later sibling runtime loop.

### Actual behavior / error logs
```text
Traceback (most recent call last):
  ...
  File "/home/zhangzhendong/ptoas-workspace/PTOAS/ptodsl/ptodsl/_ast_rewrite.py", line 634, in _rewrite_for
    raise PTODSLAstRewriteError(
ptodsl._ast_rewrite.PTODSLAstRewriteError: ast_rewrite=True runtime for-loops cannot expose the loop induction variable outside the loop yet; use explicit pto.for_(...) for 'c'
```

### Git commit
2bcde0294ab596539d5450b24a79d91b2ecb2518

### Host platform
Linux (x86_64)

### Target Ascend arch (if relevant)
a5

### PTOAS build level (if relevant)
N/A

jimmychou0:
## Summary
- analyze runtime-loop liveness from authored source AST before generating control-flow scaffolding
- preserve Python comprehension-local bindings during AST rewrite analysis
- lower direct assignment conditional expressions through merged PTODSL branches

## Validation
- rebuilt local LLVM21 PTOAS Python modules and ptoas
- ptodsl/tests/test_ast_rewrite.py
- ptodsl/tests/test_jit_compile.py
- ptodsl/tests/test_jit_diagnostics.py
- ptodsl/tests/test_docs_as_test.py
- ctest --test-dir build-local-vpto --output-on-failure -R '^ptodsl_ast_rewrite$'

Closes #933

### DIFF
--- a/ptodsl/docs/user_guide/05-control-flow.md
+++ b/ptodsl/docs/user_guide/05-control-flow.md
@@ -304,6 +304,22 @@ def ast_rewrite_side_effect_kernel():
         pto.pipe_barrier(pto.Pipe.ALL)
 ```
 
+### Conditional expressions
+
+A conditional expression used as the complete right-hand side of a normal
+(`=`) or annotated assignment is rewritten with the same device-side branch
+semantics. For example,
+`count = remaining if has_tail else 64` produces a merged runtime value when
+`has_tail` is a PTO runtime condition. A Python scalar literal branch is
+materialized using the other branch's runtime scalar type. Both branches must
+therefore provide, or allow PTODSL to infer, one compatible runtime scalar type.
+
+This rewrite does not apply to augmented assignments such as `+=`, or to a
+conditional expression nested inside a larger right-hand-side expression. Those
+forms keep normal Python tracing semantics, so their condition must be a
+trace-time Python boolean. For a runtime condition in either form, first assign
+the conditional result directly or use an explicit `pto.if_`.
+
 ### Runtime loops
 
 Native `range(...)` loops become device-side loops:
--- a/ptodsl/ptodsl/_ast_rewrite.py
+++ b/ptodsl/ptodsl/_ast_rewrite.py
@@ -156,8 +156,11 @@ class _NameInfoVisitor(ast.NodeVisitor):
     def __init__(self):
         self.loads = set()
         self.stores = set()
+        self._comprehension_bindings = set()
 
     def visit_Name(self, node):
+        if node.id in self._comprehension_bindings:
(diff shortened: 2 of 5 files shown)


Zhendong404:
issue上的这几个问题我在vmi分支上修掉了

PR #724, opened 2026-05-28
Add tmp support for tci/trowexpandadd and printFormat for tprint

FangRui0:

PR #971, opened 2026-07-21
feat(vpto): add A2/A3 gather lowering

castigli:
## Summary

- Add A2/A3 VPTO lowering for `pto.tgatherb` through `pto.ub.vgatherb` to packed `llvm.hivm.VGATHERB.b16/.b32` calls.
- Lower `pto.tgather` index form through `vmuls + vgather`, preserving count-mode mask lifetime and chunking wide rows.
- Model A2/A3 `tgatherb` as a 32-byte block gather with one compact i32 block address per output block, padded to eight entries per repeat.
- Add UB op verification/effects, synchronization modeling, design notes, lit coverage, and NPU e2e tests.
- Build on the generic PTODSL/A5 `tgatherb` support merged in #932 without duplicating its wrapper or namespace registration.

## Notes

- `tgatherb` is a 32-byte block gather, not scalar element gather.
- Scalar arbitrary-element gather is handled by the `tgather` index form and `vgather`.
- A2/A3 `mgather` is supported by PTO-ISA/EmitC; GM-to-UB VPTO lowering is intentionally split into a stacked follow-up because it requires scalar index/OOB handling and MTE2 decomposition.

## Validation

- Hardened Docker `ptoas` build: passed.
- Gather-filter lit tests: 40 passed, including all 11 PR-specific gather tests.
- 910B2 NPU e2e on devices 2 and 3: 15 passed.

Covered NPU shapes:

- `tgatherb` f32: `1x8`, `1x64`, `1x96`, `2x64`, `1x128`, `4x64`
- `tgatherb` f16: `1x16`, `1x128`, `1x192`
- scalar `tgather` f32/f16: `1x64`, `2x64`, `2x128`


lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -811,6 +811,202 @@ struct LowerPTOToUBufOpsPass
...
+        // vgatherb reads 8 u32 block addresses per repeat. Each address must
+        // point at a 32B source block; the instruction copies those 8 blocks.
+        int64_t numRepeatPerLine = validCol / epr;
+        int64_t numRemainPerLine = validCol % epr;
+        constexpr int64_t REPEAT_MAX = 255;
+        constexpr int64_t ADDRS_PER_REPEAT = 8;
+
+        // Compact offset tile layout: one i32 address per 32B output block,
+        // padded to 8 entries because even a tail repeat reads 8 addresses.
+        int64_t blocksPerRow = (validCol + bse - 1) / bse;
```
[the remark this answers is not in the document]
castigli: No change. On supported targets, int64_t represents every value of the unsigned bse, so the usual arithmetic conversion is safe and remains signed. Hardened -Werror builds also compile this expression without warnings.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -811,6 +811,202 @@ struct LowerPTOToUBufOpsPass
...
+          continue;
+
+        // Extract the source base address (i64) from the CastPtrOp that
+        // defines the src tile pointer. Only handle alloc_tile-backed src.
+        Value srcAddr;
+        if (auto *defOp = op.getSrc().getDefiningOp())
+          if (isa<pto::CastPtrOp>(defOp))
+            srcAddr = defOp->getOperand(0);
+        if (!srcAddr)
+          continue;
```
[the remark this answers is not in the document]
castigli: Addressed in 904cb4471. The A2/A3 lowering now emits a clear error and fails the pass when the index-form gather source does not provide the expected alloc-backed planned address.

lib/PTO/Transforms/VPTOLLVMEmitter.cpp
```diff
@@ -4867,6 +4867,157 @@ class LowerUBufBinaryOpPattern final : public OpConversionPattern<UBOp> {
...
+      return rewriter.create<arith::ShLIOp>(loc, v, constI64(amount));
+    };
+
+    // config[31:0] = source data address (low 32 bits of the src pointer).
+    // Trace back through castptr to get the planned UB offset, matching the
+    // address loaded from Tile host_ptr metadata by the PTO-ISA reference.
+    Value srcAddr;
+    if (auto *defOp = op.getSrc().getDefiningOp()) {
+      if (auto castOp = dyn_cast<pto::CastPtrOp>(defOp))
+        srcAddr = castOp.getOperand();
```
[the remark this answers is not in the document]
castigli: No change. CastPtrOp has MLIR OpTrait::OneOperand, which explicitly provides getOperand(). The generated named accessor is getInput(), so there is no generated getOperand hiding conflict.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -811,6 +811,202 @@ struct LowerPTOToUBufOpsPass
...
+        auto info = extractTileShapeInfo(op, tileShapes);
+        if (!info)
+          continue;
+
+        unsigned bse = info->blockSizeElem;
+        unsigned epr = info->elementsPerRepeat;
+        int64_t validRow = info->vRows;
+        int64_t validCol = info->vCols;
+        if (bse == 0 || epr == 0 || validCol == 0)
```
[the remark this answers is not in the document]
castigli: No change. Negative tile valid dimensions are canonicalized to dynamic, and this pass rejects dynamic valid shapes before lowering. Therefore validCol is nonnegative here; zero is already handled explicitly.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -811,6 +811,204 @@ struct LowerPTOToUBufOpsPass
...
+              i64c(1, loc, builder), i64c(repeat, loc, builder));
+        };
+
+        // ---- GatherBlockHead: process full epr-element blocks ----
+        if (numRepeatPerLine > 0) {
+          int64_t numLoop = numRepeatPerLine / REPEAT_MAX;
+          int64_t remainAfterLoop = numRepeatPerLine % REPEAT_MAX;
+
+          for (int64_t i = 0; i < validRow; ++i) {
+            int64_t rowOff = i * validCol;
```
zhangstevenunity, line 882: **P1: this must be the allocated row stride, not `validCol`.**

pto-isa passes `dstRowStride = TileDataDst::RowStride` and `offsetRowStride = TileDataOffset::RowStride` into `TGatherBlock`, and `pto_tile.hpp:1447` defines `RowStride = Cols` for row-major -- the allocated cols, not the valid cols. `TGatherOp::verify` pins `dstValid[1] == dstShape[1]` (PTO.cpp:7795) so its lowering is safe; `TGatherBOp::verify` has no equivalent, so `v_col < cols` reaches here.

Repro -- src/dst `rows=2, cols=128, v_row=2, v_col=64`, offsets `rows=2, cols=8, v_row=2, v_col=8`:

```
// --pto-backend=vpto, after pto-lower-to-ubuf-ops
%3 = pto.addptr %2, %c0    // row 0
pto.ub.vgatherb %3, %4, %0, %c8_i64, %c1_i64, %c1_i64
%5 = pto.addptr %2, %c64   // row 1  <-- should be %c128
pto.ub.vgatherb %5, %6, %0, %c8_i64_2, %c1_i64_3, %c1_i64_4
```

The EmitC path on the same input emits `Tile<TileType::Vec, float, 2, 128, BLayout::RowMajor, 2, 64, ...>` and `TGATHERB(v10, v4, v7)`, so pto-isa uses stride 128. Two backends, one input, different results -- and row 1 of the output is never written.

Same for the offsets pointer: `offsetStridePerRow` is computed from `validCol` instead of read off the offsets tile, so an offsets tile allocated `2x16` with valid `2x8` (which the new verifier accepts) advances by 8 here and by 16 in pto-isa.

Suggest taking both strides from `info->cols` / a second `extractTileShapeInfoFromValue(op.getOffsets(), ...)`, and adding a lit case with `v_col < cols`.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
```diff
@@ -811,6 +811,204 @@ struct LowerPTOToUBufOpsPass
...
+    // pto-isa a2a3/TGather.hpp.
+    {
+      SmallVector<pto::TGatherOp> ops;
+      func.walk([&](pto::TGatherOp op) { ops.push_back(op); });
+      for (auto op : ops) {
+        if (!op.hasIndexForm())
+          continue;
+        if (!canLower(op, tileShapes))
+          continue;
+        auto info = extractTileShapeInfo(op, tileShapes);
```
zhangstevenunity, line 941: **P1: `extractTileShapeInfo` returns the *indices* tile here, not `dst`.**

`extractTileShapeInfo` is hard-coded to `op->getOperand(2)` (line 205). For the elementwise ops it was written for that is `dst`, but `TGatherOp` is `(src, dst, [cdst], [indices], [tmp])`, so in index form operand 2 is `indices`. Everything below -- `info->vRows`, `info->vCols`, `info->cols` (used for the `dst` pointer at line 985) and `epr` -- is therefore taken off the index tile.

`TGatherOp::verify` never relates `dst` to `indices`, so this verifies clean:

```mlir
%src     = ... f32, rows=1, cols=64,  v_row=1, v_col=64
%indices = ... i32, rows=1, cols=128, v_row=1, v_col=128
%tmp     = ... i32, rows=1, cols=128, v_row=1, v_col=128
%dst     = ... f32, rows=1, cols=64,  v_row=1, v_col=64
pto.tgather ins(%src, %indices, %tmp : ...) outs(%dst : ...)
```

and the pass emits two chunks, writing `dst+0..63` then `dst+64..127` -- 256 bytes past the end of a 256-byte dst tile. pto-isa loops on `dst.GetValidRow()/GetValidCol()` and keeps `TShape1 = TileDataD::Cols` (dst) separate from `idx_stride = TileDataS1::Cols` (indices), so it writes exactly 64 elements.

Second effect, this one on well-formed input: `elementsPerRepeat` is always `256/4 = 64` because indices are i32. On `1x128xf16` this pass emits two `set_mask 64` + two `VGATHER.b16` where pto-isa emits a single `set_vector_mask(0, 128)` + one `vgather` -- 2x the instructions for every b16 gather.

Fix: use `extractTileShapeInfoFromValue(op.getDst(), tileShapes)` for the bounds/dst stride and a separate one for `indices`/`tmp`; either tighten `verifyIndexForm` to require dst and indices to agree on valid shape, or thread the two strides through independently as pto-isa does.

zhangstevenunity (changes requested):
Deep correctness review. I built this branch (ptoas 0.55, LLVM21) and compared the emitted IR against the pto-isa a2a3 reference (`npu/a2a3/TGatherB.hpp`, `npu/a2a3/TGather.hpp`) and against this repo's own EmitC path for byte-identical input.

The design work is careful and mostly lands: the bisheng-probed VGATHERB config layout, the 32B-block-gather writeup, the InsertSync macro model, and the intrinsic selection (`VMULS.s32` / `VGATHER.b16` / `VGATHERB.b32`) all check out. But two of the tile->UB loop parameters are read from the wrong place, and both silently miscompile input that the verifier accepts.

## P1-1: tgatherb uses the tile's *valid* cols as the row stride; pto-isa uses the *allocated* cols

`LowerPTOToUBufOps.cpp:882` uses `rowOff = i * validCol` (and `offRowOff = i * offsetStridePerRow`, derived from validCol rather than the offsets tile's cols). pto-isa's `TGATHERB_IMPL` passes `dstRowStride = TileDataDst::RowStride` and `offsetRowStride = TileDataOffset::RowStride`, and `pto_tile.hpp:1447` defines `RowStride = Cols` for row-major -- i.e. the *allocated* cols.

Unlike `TGatherOp::verify`, which pins `dstValid[1] == dstShape[1]` (PTO.cpp:7795), the tgatherb verifier places no such constraint, so `v_col < cols` is legal here.

A/B on a `2x128` alloc with `v=2x64` (src+dst), offsets `2x8`:

- vpto emits row 1 at `dst + 64`
- EmitC emits `Tile<TileType::Vec, float, 2, 128, RowMajor, 2, 64, ...>` + `TGATHERB(...)`, so pto-isa uses `dst + 128`

Row 1 of the output is never written; row 0's padding is corrupted instead. Same divergence on the offsets pointer: with offsets allocated `2x16` / valid `2x8` (which the new verifier requires), vpto advances by 8 and pto-isa by 16.

## P1-2: the tgather index-form loop bounds come from `indices`, not `dst`

`extractTileShapeInfo` (LowerPTOToUBufOps.cpp:205) is hard-coded to `op->getOperand(2)`. That is `dst` for the elementwise ops it was written for, but `TGatherOp` is `(src, dst, [cdst], [indices], [tmp])`, so in index form operand 2 is `indices`. `vRows`, `vCols`, `cols` and `elementsPerRepeat` are therefore all taken off the index tile.

`TGatherOp::verify` never relates `dst` to `indices`, so `dst = 1x64xf32` with `indices/tmp = 1x128xi32` verifies clean. The lowering then emits two chunks and writes `dst+0..63` and `dst+64..127` -- 256 bytes past the end of a 256-byte dst tile. pto-isa loops on `dst.GetValidRow()/GetValidCol()` and keeps `TShape1 = TileDataD::Cols` separate from `idx_stride = TileDataS1::Cols`, so it writes exactly 64 elements.

Even on well-formed input this costs 2x the instructions for b16: `elementsPerRepeat` is always `256/4 = 64` (indices are i32). On `1x128xf16` the pass emits two `set_mask 64` + two `VGATHER.b16` where pto-isa emits one `set_vector_mask(0, 128)` + one `vgather`.

## Test quality (mutation-tested)

I replaced both `rowOff` and `offRowOff` with a literal `0` -- every row of a multi-row tgatherb gathering into row 0 from the same addresses -- rebuilt, and re-ran the gather suite: **47/47 still pass**, including `tgatherb_multi_repeat.pto`, which exists specifically to cover the 2-row case. All five lowering tests are presence-only (`CHECK: pto.ub.vgatherb` / `CHECK-NOT: pto.tgatherb`); none pins an operand, a count, a stride, or a pointer offset, so no row-stride or repeat error is detectable. `CHECK-COUNT-N` plus `CHECK-DAG` on the `pto.addptr` offsets would have caught both P1s.

## Cross-layer (per .claude/rules/cross-layer-sync.md)

The PR changes the user-visible tgatherb contract (per-element byte offsets -> one 32B block address per output block, padded to 8) and adds verifier rules for it, but both user-facing docs still describe the old semantics and none of the new constraints:

- `docs/PTO_IR_manual.md:7597` -- "Gathers elements using per-element byte offsets", `dst[i, j] = src[byte_offsets[i, j]]`
- `ptodsl/docs/user_guide/08-compute-operations.md:251` -- same

The design doc records that this exact confusion is what made the first NPU run look wrong, so leaving the manual saying the old thing seems worth fixing here.

## Verified sound

- VGATHERB/VGATHER config packing matches the documented bisheng probe; `.b16`/`.b32` suffix selection is right.
- `getTGatherBSyncMacroModel` correctly models a single PIPE_V phase reading src+offsets and writing dst; InsertSync then places the MTE2->V flags.
- `UBVmulSOp` argument order matches `vmuls(dst, src, scalar, 1, 1, 1, 8, 8)`; count-mode mask is held across vmuls/vgather and restored to norm+full at the end, matching TGather.hpp.
- Head/tail REPEAT_MAX chunking and the `j * 8 * REPEAT_MAX` offset advance match GatherBlockHead exactly.
- Gemini's four comments: the `int64_t`/`unsigned` and `getOperand()` replies are correct as written; no action needed there.
- Full lit run on this branch: the 130 failures I see are all `--enable-tile-op-expand` / tilelang paths that need a staged PTODSL package my local build does not have. No gather test fails, and CI is green.

lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
```diff
@@ -409,6 +409,17 @@ std::optional<SyncMacroModel> getMGatherSyncMacroModel(pto::MGatherOp op) {
   return model;
 }
 
+// TGatherBOp (byte-offset gather): single-phase PIPE_V op that reads src +
+// offsets and writes dst. Explicit model ensures InsertSync detects the
+// cross-pipe MTE2→V dependency (DMA load of src/offsets must complete before
+// the gather reads them).
+std::optional<SyncMacroModel> getTGatherBSyncMacroModel(pto::TGatherBOp op) {
```
zhangstevenunity, line 416: **P3: this macro model looks redundant -- the generic fallback already emits the identical sync.**

`PTOIRTranslator::RecursionIR` dispatches macro models *before* the generic handler:

```cpp
else if (getSyncMacroModel(op))         { UpdateMacroOpInfo(op); }
...
else if (isa<pto::OpPipeInterface>(op)) { UpdatePTOOpInfo(op); }   // fallback
```

`UpdatePTOOpInfo` takes the pipe from `OpPipeInterface::getPipe()` and the def/use sets straight from `MemoryEffectOpInterface::getEffects()`. `TGatherBOp` already had both before this PR:

- `getPipe() { return PIPE_V; }` -- PTOOps.td:5182
- `PTO_DEFINE_BINARY_EFFECTS(TGatherBOp, getSrcMutable(), getOffsetsMutable(), getDstMutable())` -- PTO.cpp:14732

which is exactly what this model declares (single PIPE_V phase, uses = {src, offsets}, defs = {dst}).

A/B: I deleted the `TGatherBOp` dispatch from `getSyncMacroModel`, rebuilt, and compiled a `TLOAD(src) / TLOAD(offsets) -> tgatherb -> TSTORE` kernel with `--pto-arch=a3 --pto-level=level3 --enable-insert-sync`. The emitted synchronization is byte-identical with and without this model:

```
TLOAD(v11, v21);
set_flag(PIPE_MTE2, PIPE_V, EVENT_ID0);
wait_flag(PIPE_MTE2, PIPE_V, EVENT_ID0);
TGATHERB(v14, v8, v11);
set_flag(PIPE_V, PIPE_MTE3, EVENT_ID0);
wait_flag(PIPE_V, PIPE_MTE3, EVENT_ID0);
TSTORE(v24, v14);
```

The entire diff between the two builds is pointer-derived type-name hashes (`GT_160059936` vs `GT_1013768560`).

So the design doc's "InsertSync did not insert MTE2->V sync around TGatherBOp because it lacked a sync macro model" does not reproduce at this head. If you still have the failing case from that debugging session, it is worth posting -- it would mean the real cause is elsewhere (effects, or the alias analysis) and this model is only masking it.

Two reasons I would not just leave it in:

1. The macro-model branch *replaces* the generic path rather than supplementing it, so the operand list is now duplicated in a second place. If anyone later adds an operand to `TGatherBOp::getEffects` (a tmp scratch, say), the generic path would pick it up automatically and this model would silently not -- the same shape of bug as #724.
2. The design doc's root-cause note will send the next person debugging a gather sync issue down the wrong path.

Incidentally, this also answers "why was the EmitC path not updated too": `PTOInsertSyncPass` runs on the shared mainline at ptoas.cpp:3253, before `PTOMaterializeTileHandles` (:3294, "so both backends consume the same post-planning seam IR") and before any `effectiveBackend == PTOBackend::EmitC` branch. The transcript above is EmitC output. There is only one sync path, so a change here is never vpto-only -- worth keeping in mind for the verifier tightening in `TGatherBOp::verify` as well, which is arch-gated rather than backend-gated.

zhangstevenunity (commented):
Follow-up to my earlier review, on the InsertSync part specifically. I A/B-tested it and it appears to be a no-op -- the generic `OpPipeInterface` + `MemoryEffects` fallback already produces the identical sync. Details inline. Not a bug, but the design doc records a root cause that does not hold at this head, which is worth resolving before merge.

castigli:
Thanks for the thorough review. Addressed in `d5db12603`.

For P1-1, `tgatherb` now extracts destination and offsets metadata independently and uses their allocated physical row strides. The tail repeat stride also comes from the allocated destination width rather than `validCol`.

For P1-2, index-form `tgather` now derives its loop bounds and elements-per-repeat from `dst`. Destination, indices, and tmp pointers each advance using their own physical row stride. The A2/A3 verifier now requires matching destination/index valid shapes and row-major layouts.

I also strengthened the coverage:

- Added a padded multi-row `tgatherb` regression that pins destination offsets `128/192`, offsets-tile offsets `24/32`, and exactly four `vgatherb` calls.
- Added an f16 index-gather regression covering destination-derived 128-lane operation and independent tmp/index/dst strides.
- Added negative verifier tests for destination/index shape mismatch, unsupported offsets layout, and unsupported A2/A3 b8 `tgatherb`.
- Updated the PTO IR manual, PTODSL guide, operation summary, and API documentation to describe the 32-byte block-address contract.

For the InsertSync follow-up, confirmed the dedicated `TGatherBOp` macro model was redundant with the existing `OpPipeInterface` + `MemoryEffectsOpInterface` path. I removed it, corrected the design note, and added a regression proving the generic path emits the MTE2→V synchronization.

Validation using the LLVM 21/CANN 9 Docker image through `task-submit`:

- Docker rebuild: passed
- Focused review regressions: 4/4 passed
- Gather-filter lit suite: 49/49 passed
- Full lit suite: 1,442 passed, 1 unsupported, 0 failed
- A3 VPTO NPU gather E2E: 15/15 passed

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
castigli: Addressed in the current rebased series (`211cb3bc3`). `tgatherb` now extracts destination and offsets metadata independently and advances rows with their allocated `cols` strides. The padded multi-row regression pins destination offsets 128/192 and offsets-tile offsets 24/32.

lib/PTO/Transforms/InsertSync/SyncMacroModel.cpp
castigli: Addressed in the current rebased series (`211cb3bc3`). The dedicated `TGatherBOp` macro model and dispatch were removed, the design explanation was corrected, and `tgatherb_generic_sync.pto` verifies that the generic `OpPipeInterface` plus memory-effects path emits the MTE2-to-V synchronization.

lib/PTO/Transforms/LowerPTOToUBufOps.cpp
castigli: Addressed in the current rebased series (`211cb3bc3`). Index-form `tgather` now derives bounds and elements-per-repeat from `dst`, extracts `indices` and `tmp` metadata separately, and advances each pointer with its own physical stride. The A2/A3 verifier requires matching destination/index valid shapes and row-major destination/index/tmp layouts; the f16 regression covers the 128-lane case and independent strides.

castigli:
Re-review update:

- Rebased linearly onto current `main` (`1a4d96961`, LLVM 19 permanent baseline); current head is `0096c8327`.
- Confirmed both P1 correctness fixes remain intact and removed the redundant `TGatherBOp` sync macro model.
- Added braces to all PR-introduced gather control flow per the current compliance rules.
- Clean `ptoas:dev-cann900-llvm-fork` build passed against LLVM 19.1.7 through `task-submit`.
- Gather-filtered lit suite passed: 61/61.
- GitHub build/test, simulator, license, source-patch, and both wheel jobs all passed.
- Replied to and resolved all three outstanding review threads.

Please re-review when convenient.

castigli:
@zhangstevenunity all requested changes are addressed, the outstanding threads are resolved, and the rebased LLVM 19 CI is fully green. Please re-review when convenient.

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`e19aff7dda7a`
- 结果汇总：OK `320` / FAIL `2` / SKIP `30`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260825_054632_merged_pr971.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260825_054632_merged_pr971.tsv`

### 失败用例
- `Ci/ci` (run, exit=2)
- `Qwen3_14BPrefillA3/out_proj_aiv` (run, exit=2)

Issue #995, opened 2026-07-24
[Bug] 0.51 lowers cmo.cacheinvalid on a partition_tensor_view to an uncompilable DCCI call


luohuan19:
### Component

EmitC / Codegen (lib/PTO/Transforms/PTOToEmitC.cpp)

### Description

Since 0.51, `pto.cmo.cacheinvalid <partition_tensor_view> single_cache_line` is
lowered to a call that PTOAS's own emitted helper cannot compile.

The helper PTOAS emits is pointer-only:

```cpp
template <typename Ptr>
static AICORE inline void PTOAS__DCCI_SINGLE_CACHE_LINE(Ptr ptr) {
  dcci((__gm__ void*)ptr, cache_line_t::SINGLE_CACHE_LINE);
}
```

but for a `partition_tensor_view` operand the argument passed in is the
materialised `GlobalTensor<...>` object, not its address. `GlobalTensor` has no
conversion to `__gm__ void*`, so every kernel carrying this op fails to compile.

I checked pto-isa at both the pinned commit `83d01313` and current `main`
(`439faf48`): `GlobalTensor` exposes `data()` and `SetAddr()` and has no
conversion operator to `__gm__ void*` in either, so this is not a pto-isa version
skew — no pto-isa revision makes the emitted cast valid.

0.50 also accepted this IR but emitted **no** call at all for it (only the helper
template definition), so the op was silently a no-op. Either behaviour is a
problem, but the 0.51 one is a hard build break.

Impact on the PyPTO side: this takes down 36 `dist-system-tests` (all L3
collectives) and the DeepSeek-V4-Flash `moe` model example when moving the
toolchain pin from 0.50 to 0.51. PyPTO's `InsertCommFence` pass emits this op to
implement the publish side of the data-before-signal contract; we have had to
suspend emitting it entirely to get off 0.50.

### Reproduction (minimal)

`repro.pto`:

```mlir
module attributes {pto.target_arch = "a2a3"} {
  func.func @k(%arg0: !pto.ptr<f32>) attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
  %c16_index = arith.constant 16 : index
  %c1_index = arith.constant 1 : index
  %c0_index = arith.constant 0 : index
  %view = pto.make_tensor_view %arg0, shape = [%c16_index, %c16_index], strides = [%c16_index, %c1_index] {layout = #pto.layout<nd>}: !pto.tensor_view<?x?xf32>
[...]

PR #1008, opened 2026-07-27
Support odd-axis split modes for tpush/tpop

zhangstevenunity:
## Summary
- accept split modes 3 and 4 for tpush, tpop, and tfree verification
- emit TILE_UP_DOWN_ODD and TILE_LEFT_RIGHT_ODD PTO-ISA template arguments
- update pipe split inference, documentation, and regression coverage

## Validation
- WSL LLVM 21 build: cmake --build build-codex-odd-split -j1 --target ptoas_runtime_staging
- llvm-lit --filter=tpush_tpop_: 34 passed
- git diff --cached --check

### DIFF
--- a/lib/PTO/IR/PTO.cpp
+++ b/lib/PTO/IR/PTO.cpp
@@ -15231,8 +15231,8 @@ static bool isInsideSectionOrAttributedKernel(Operation *op) {
 }
 
 static LogicalResult verifySplitAttr(Operation *op, int64_t split) {
-  if (split < 0 || split > 2)
-    return op->emitOpError("expects 'split' to be 0, 1, or 2");
+  if (split < 0 || split > 4)
+    return op->emitOpError("expects 'split' to be 0, 1, 2, 3, or 4");
   return success();
 }
 
@@ -16094,7 +16094,7 @@ static LogicalResult verifyAivSubblockIdOperand(Operation *op,
 
   if (split == 0) {
     return op->emitOpError(
-        "expects 'aiv_subblockid' only when 'split' is 1 or 2");
+        "expects 'aiv_subblockid' only when 'split' is 1, 2, 3, or 4");
   }
 
   if (isa<TensorViewType>(pipeEntryType)) {
--- a/lib/PTO/Transforms/PTOToEmitC.cpp
+++ b/lib/PTO/Transforms/PTOToEmitC.cpp
@@ -1114,6 +1114,10 @@ static FailureOr<std::string> getTileSplitToken(int64_t split) {
     return std::string("TileSplitAxis::TILE_UP_DOWN");
   case 2:
     return std::string("TileSplitAxis::TILE_LEFT_RIGHT");
+  case 3:
+    return std::string("TileSplitAxis::TILE_UP_DOWN_ODD");
+  case 4:
+    return std::string("TileSplitAxis::TILE_LEFT_RIGHT_ODD");
   default:
     return failure();
   }
--- a/test/lit/pto/tpush_tpop_odd_split_emitc.pto
+++ b/test/lit/pto/tpush_tpop_odd_split_emitc.pto
@@ -1,0 +1,84 @@
+// Copyright (c) 2026 Huawei Technologies Co., Ltd.
+// This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+// CANN Open Software License Agreement Version 2.0 (the "License").
+// Please refer to the License for details. You may not use this file except in compliance with the License.
+// THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+// See LICENSE in the root of the software repository for the full text of the License.
+//
+// Verifies that odd valid-shape split modes survive frontend pipe lowering and
+// select the matching PTO-ISA template arguments for both TPUSH and TPOP.
+//
+// RUN: ptoas --pto-arch=a5 %s 2>&1 | FileCheck %s --check-prefix=A5
+
+module {
+  func.func @cube_kernel(%full_rows: index, %full_cols: index)
+      attributes {pto.kernel_kind = #pto.kernel_kind<cube>} {
+    %c16 = arith.constant 16 : index
+
+    %up_down_import = pto.import_reserved_buffer {
+      name = "up_down_fifo",
+      peer_func = @vector_kernel
+    } -> i32
+    %left_right_import = pto.import_reserved_buffer {
+      name = "left_right_fifo",
+      peer_func = @vector_kernel
+    } -> i32
+
+    pto.aic_initialize_pipe {id = 3, dir_mask = 1, slot_size = 1024}
+      (c2v_consumer_buf = %up_down_import : i32)
+    pto.aic_initialize_pipe {id = 4, dir_mask = 1, slot_size = 1024}
+      (c2v_consumer_buf = %left_right_import : i32)
+
+    %up_down_tile = pto.alloc_tile valid_row = %full_rows valid_col = %c16
+      : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>
+    %left_right_tile = pto.alloc_tile valid_row = %c16 valid_col = %full_cols
+      : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>
+
+    pto.tpush_to_aiv(%up_down_tile : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>) {id = 3, split = 3}
+    pto.tpush_to_aiv(%left_right_tile : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>) {id = 4, split = 4}
+    return
+  }
+
+  func.func @vector_kernel(%half_rows: index, %half_cols: index)
+      attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
+    %c16 = arith.constant 16 : index
+
+    %up_down_local = pto.reserve_buffer {
+      name = "up_down_fifo",
+      size = 8192,
+      location = #pto.address_space<vec>,
+      auto = true
+    } -> i32
+    %left_right_local = pto.reserve_buffer {
+      name = "left_right_fifo",
+      size = 8192,
+      location = #pto.address_space<vec>,
+      auto = true
+    } -> i32
+
+    pto.aiv_initialize_pipe {id = 3, dir_mask = 1, slot_size = 1024}
+      (c2v_consumer_buf = %up_down_local : i32)
+    pto.aiv_initialize_pipe {id = 4, dir_mask = 1, slot_size = 1024}
+      (c2v_consumer_buf = %left_right_local : i32)
+
+    %up_down_tile = pto.tpop_from_aic(%half_rows, %c16) {id = 3, split = 3}
+      -> !pto.tile_buf<loc=vec, dtype=f32, rows=8, cols=16, v_row=?, v_col=?, blayout=row_major, slayout=none_box, fractal=512, pad=0>
+    pto.tfree_from_aic {id = 3, split = 3}
+
+    %left_right_tile = pto.tpop_from_aic(%c16, %half_cols) {id = 4, split = 4}
+      -> !pto.tile_buf<loc=vec, dtype=f32, rows=16, cols=8, v_row=?, v_col=?, blayout=row_major, slayout=none_box, fractal=512, pad=0>
+    pto.tfree_from_aic {id = 4, split = 4}
+    return
+  }
+}
+
+// A5-LABEL: AICORE void cube_kernel(
+// A5: TPUSH<{{.*}}TileSplitAxis::TILE_UP_DOWN_ODD>(
+// A5: TPUSH<{{.*}}TileSplitAxis::TILE_LEFT_RIGHT_ODD>(
+
+// A5-LABEL: AICORE void vector_kernel(
+// A5: TPOP<{{.*}}TileSplitAxis::TILE_UP_DOWN_ODD>(
+// A5: TFREE<{{.*}}TileSplitAxis::TILE_UP_DOWN_ODD>(
+// A5: TPOP<{{.*}}TileSplitAxis::TILE_LEFT_RIGHT_ODD>(
+// A5: TFREE<{{.*}}TileSplitAxis::TILE_LEFT_RIGHT_ODD>(
--- a/docs/PTO_IR_manual.md
+++ b/docs/PTO_IR_manual.md
@@ -9206,8 +9206,11 @@ frontend/framework generated IR. The detailed design document is:
 - `nosplit` is an optional compile-time boolean attribute on
   `pto.aic_initialize_pipe` / `pto.aiv_initialize_pipe`.
 - `split` is a compile-time attribute, not a runtime SSA operand.
-- `split = 0/1/2` corresponds to `TILE_NO_SPLIT`, `TILE_UP_DOWN`, and
-  `TILE_LEFT_RIGHT`.
+- `split = 0/1/2/3/4` corresponds to `TILE_NO_SPLIT`, `TILE_UP_DOWN`,
+  `TILE_LEFT_RIGHT`, `TILE_UP_DOWN_ODD`, and `TILE_LEFT_RIGHT_ODD`.
+- `TILE_UP_DOWN_ODD` splits an odd valid-row count so AIV0 receives
+  `ceil(valid_rows / 2)` rows and AIV1 receives `floor(valid_rows / 2)` rows.
+  `TILE_LEFT_RIGHT_ODD` applies the same rule to valid columns.
 - `pto.tpop_from_aic` and `pto.tpop_from_aiv` are result-valued frontend ops.
 - Pipe entries support two forms:
   - tile entry: `!pto.tile_buf<...>` or the equivalent local memref after view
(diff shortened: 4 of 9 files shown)


lib/PTO/IR/PTO.cpp
```diff
@@ -15231,8 +15231,8 @@ static bool isInsideSectionOrAttributedKernel(Operation *op) {
 }
 
 static LogicalResult verifySplitAttr(Operation *op, int64_t split) {
-  if (split < 0 || split > 2)
-    return op->emitOpError("expects 'split' to be 0, 1, or 2");
+  if (split < 0 || split > 4)
```
zhangstevenunity, line 15234: The range now admits 3/4, but nothing downstream constrains *where* they are legal.

pto-isa `505dbfa6` (the commit that introduced the ODD enumerators) implements them in exactly one function per arch, `popVecTileFromGMFiFo`. Every producer helper is untouched and none has a catch-all that behaves correctly:

- a5 `pushAcc2VecFiFo` (C2V UB local FIFO) and `pushVec2MatFiFo` (V2C L1): the `if constexpr` chain is NO_SPLIT / UP_DOWN / LEFT_RIGHT with no `else`. With `Split == TILE_UP_DOWN_ODD` no branch matches, so no `TMOV_IMPL` / `TINSERT_IMPL` is emitted at all and the push is a silent no-op.
- a5 `pushVec2GMFiFo` / a2a3 `pushVec2GMFiFo` (V2C GM): terminate in `} else { // TILE_LEFT_RIGHT`, so both ODD modes silently take the *even* left-right offset `subBlockId * validC * sizeof(T)` and the unsplit `gmStrideR`.

So the only end-to-end correct configuration is C2V through a GM FIFO (cube writes the whole slot via `pushAcc2GMFiFo`, which takes no `Split`, and each AIV pops its odd half).

Suggest restricting 3/4 here to that configuration with an actionable diagnostic - `verifySplitAttr` has the `Operation *` and can reach the matched init op through `lookupFrontendInitOpById` - rather than accepting them on every talloc/tpush/tpop/tfree.

Separately: there is no parity check. `split = 3` on a statically even `v_row` (and `split = 1` on a statically odd one) is accepted silently, even though pto-isa's contract is "row axis must be odd number". The even local paths at least carry `static_assert(ProdM % 2 == 0)`; the GM path has no assert at all, so a mismatch is silent wrong-address arithmetic. Where `v_row` / `v_col` are static this is cheap to catch here.

lib/PTO/Transforms/PTOToEmitC.cpp
```diff
@@ -1114,6 +1114,10 @@ static FailureOr<std::string> getTileSplitToken(int64_t split) {
     return std::string("TileSplitAxis::TILE_UP_DOWN");
   case 2:
     return std::string("TileSplitAxis::TILE_LEFT_RIGHT");
+  case 3:
+    return std::string("TileSplitAxis::TILE_UP_DOWN_ODD");
```
zhangstevenunity, line 1118: These tokens do not exist in the pto-isa revision this repo pins.

`TileSplitAxis` at the pinned commit `9cf767cb` (2026-07-08, `include/pto/common/fifo.hpp`) is still:

```cpp
enum TileSplitAxis : uint8_t
{
    TILE_NO_SPLIT = 0,
    TILE_UP_DOWN = 1,
    TILE_LEFT_RIGHT = 2,
};
```

The ODD enumerators arrived in pto-isa `505dbfa6` ("Udpate tpush/tpop to support odd-number axis split", 2026-07-27), which is newer than the pin.

That pin drives every compile-based path in the repo: `.github/workflows/ci.yml:52` and `:406`, `docker/Dockerfile:137`, `test/npu_validation/scripts/run_remote_npu_validation.sh:17`, and `docs/no_npu_compile_only_guide_zh.md:171,223`. Any kernel emitting `split = 3/4` will fail to compile with "no member named 'TILE_UP_DOWN_ODD' in 'pto::TileSplitAxis'" - same failure class as #995 / PR #1001.

The lit tests only FileCheck the emitted text and never compile it, so CI stays green.

This PR should bump the pin to `>= 505dbfa6` in all five locations (`.github/scripts/update_pto_isa_pin.py` handles it).

test/lit/pto/tpush_tpop_odd_split_emitc.pto
```diff
@@ -0,0 +1,84 @@
...
+      (c2v_consumer_buf = %up_down_import : i32)
+    pto.aic_initialize_pipe {id = 4, dir_mask = 1, slot_size = 1024}
+      (c2v_consumer_buf = %left_right_import : i32)
+
+    %up_down_tile = pto.alloc_tile valid_row = %full_rows valid_col = %c16
+      : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>
+    %left_right_tile = pto.alloc_tile valid_row = %c16 valid_col = %full_cols
+      : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>
+
+    pto.tpush_to_aiv(%up_down_tile : !pto.tile_buf<loc=acc, dtype=f32, rows=16, cols=16, v_row=?, v_col=?, blayout=col_major, slayout=row_major, fractal=1024, pad=0>) {id = 3, split = 3}
```
zhangstevenunity, line 38: This test pins down the one configuration that is known-broken in pto-isa.

`pto.aic_initialize_pipe` here has only `c2v_consumer_buf` and no `gm_slot_tensor`, so it is an l2l pipe and `getTPipeDirectionToken` (`lib/PTO/Transforms/PTOToEmitC.cpp:1129`) emits `Direction::DIR_C2V`, not `DIR_C2V_GM`. On a5 that selects the local UB FIFO, and pushing a `loc=acc` tile through it lands in pto-isa's `pushAcc2VecFiFo`:

```cpp
if constexpr (Split == TileSplitAxis::TILE_NO_SPLIT) { ...
} else if constexpr (Split == TileSplitAxis::TILE_UP_DOWN) { ...
} else if constexpr (Split == TileSplitAxis::TILE_LEFT_RIGHT) { ... }
// no else
```

With `Split == TILE_UP_DOWN_ODD` no branch matches, so no `TMOV_IMPL` is emitted and the TPUSH moves nothing. It compiles cleanly, and because the FileCheck only asserts the template-argument text, the test passes.

To actually cover the supported path, the pipe should carry a `gm_slot_tensor` so it lowers to `DIR_C2V_GM` and the consumer goes through `popVecTileFromGMFiFo`, which is the only helper `505dbfa6` taught about ODD.

Also worth a comment in the test: with ODD the two sub-cores must pass *different* valid shapes to `tpop` (AIV0 `ceil`, AIV1 `floor`). `popVecTileFromGMFiFo` derives both `subAIVOffset` and `gmStrideR` from the consumer's own `tile.GetValidRow()/GetValidCol()`, so the identities only hold under that convention. `%half_rows` / `%half_cols` being opaque function arguments hides that requirement rather than demonstrating it.

zhangstevenunity (commented):
## Review: odd-axis split modes for tpush/tpop

The PTOAS-side change is mechanically correct and minimal: `verifySplitAttr` range,
`getTileSplitToken` cases 3/4, the mixed-split diagnostic text, and the docs all line
up with each other, and nothing in `lib/` or `include/` still assumes `split <= 2`
(the only remaining comparisons are `getSplit() != 0`, which stay correct).

However, I do not think this can be merged as-is: the generated code cannot compile
against the pto-isa revision this repo pins, and at pto-isa HEAD only **one** of the
transfer paths actually implements the ODD modes, while this PR opens `split = 3/4`
on all of them. The new lit test happens to exercise one of the broken paths.

The lit tests are text-only FileChecks over emitted C++, so neither problem is
visible in CI.

---

### P1-1. The pinned pto-isa does not have `TILE_UP_DOWN_ODD` / `TILE_LEFT_RIGHT_ODD`

`TileSplitAxis` at the pinned commit `9cf767cb` (2026-07-08, `include/pto/common/fifo.hpp`):

```cpp
enum TileSplitAxis : uint8_t
{
    TILE_NO_SPLIT = 0,
    TILE_UP_DOWN = 1,
    TILE_LEFT_RIGHT = 2,
};
```

The ODD enumerators were added by pto-isa `505dbfa6` ("Udpate tpush/tpop to support
odd-number axis split", 2026-07-27), which is *newer* than the pin.

That pin is what every compile-based path in this repo uses:

- `.github/workflows/ci.yml:52` and `:406`
- `docker/Dockerfile:137`
- `test/npu_validation/scripts/run_remote_npu_validation.sh:17`
- `docs/no_npu_compile_only_guide_zh.md:171,223`

So any kernel that uses `split = 3/4` emits `TileSplitAxis::TILE_UP_DOWN_ODD` and
fails to compile with "no member named 'TILE_UP_DOWN_ODD' in 'pto::TileSplitAxis'".
This is the same failure class as #995 / PR #1001.

The PR needs to bump the pin to `>= 505dbfa6` in all five locations
(`.github/scripts/update_pto_isa_pin.py` does this).

### P1-2. pto-isa implements ODD only on the C2V **GM** pop path; this PR opens it everywhere

pto-isa `505dbfa6` touches exactly one function per arch: `popVecTileFromGMFiFo`
(a2a3 `TPush.hpp:370`, a5 `TPush.hpp:584`). Every producer-side helper is untouched,
and none of them has a defaulted branch that would do the right thing:

| pto-isa helper (a5) | direction | `Split == *_ODD` behavior |
|---|---|---|
| `popVecTileFromGMFiFo` | C2V GM | correct, this is the one implemented case |
| `pushAcc2VecFiFo` (`TPush.hpp:197`) | C2V UB (local FIFO) | `if constexpr` chain is NO_SPLIT / UP_DOWN / LEFT_RIGHT with **no** `else`. ODD matches nothing, so **no `TMOV_IMPL` is emitted at all** and the push is a silent no-op. |
| `pushVec2MatFiFo` (`TPush.hpp:239`) | V2C L1 | same shape, no `TINSERT_IMPL` emitted, silent no-op |
| `pushVec2GMFiFo` (a5 `:288`, a2a3 `:165`) | V2C GM | terminates in `} else { // TILE_LEFT_RIGHT`, so both ODD modes silently get the *even* LEFT_RIGHT offset `subBlockId * validC * sizeof(T)` and the unsplit `gmStrideR` |

Only the C2V-through-GM configuration works end to end (cube writes the whole slot
through `pushAcc2GMFiFo`, which takes no `Split` at all, and each AIV pops its odd
half). On a2a3 that is the only C2V shape, so a2a3 C2V is fine. On a5 it requires an
`l2g2l` / `gm_slot_tensor` pipe so that `getTPipeDirectionToken`
(`lib/PTO/Transforms/PTOToEmitC.cpp:1129`) selects `DIR_C2V_GM`.

The new test `test/lit/pto/tpush_tpop_odd_split_emitc.pto` uses
`pto.aic_initialize_pipe` with only `c2v_consumer_buf` and no `gm_slot_tensor`, so it
lowers to `Direction::DIR_C2V` (local UB FIFO) and pushes from an `loc=acc` tile.
That is `pushAcc2VecFiFo` exactly, i.e. the case where the ODD push compiles cleanly
and moves nothing. The FileCheck only asserts the template argument text, so it
passes while locking in the configuration that silently drops data.

Suggested resolution, in order of preference:

1. Reject `split = 3/4` in the verifier for everything except the C2V GM-backed pop
   configuration pto-isa implements, with an actionable message, and rewrite the new
   test to use a `gm_slot_tensor` pipe. This keeps the feature honest and is a small
   addition to `verifySplitAttr` (it already has the op and can reach the matched
   init op via `lookupFrontendInitOpById`).
2. Or hold the PR until the pto-isa push side handles ODD, and land both together.

### P2-1. No parity validation

pto-isa's contract is explicit: `TILE_UP_DOWN_ODD` "row axis must be odd number",
`TILE_LEFT_RIGHT_ODD` "column axis must be odd number". PTOAS accepts `split = 3` on
a tile with a statically **even** `v_row`, and `split = 1` on a statically odd one,
with no diagnostic. For the even modes pto-isa at least has
`static_assert(ProdM % 2 == 0)` / `PTO_ASSERT` on the local paths, but the GM path
(the one that matters here) has neither, so a mode/parity mismatch is silent
wrong-address arithmetic.

Where `v_row` / `v_col` are static, this is cheap to check in `verifySplitAttr` and
would catch the most likely user error for a feature whose entire purpose is parity.

### P2-2. Manual does not say the two sub-cores must pass *different* valid shapes

`docs/PTO_IR_manual.md:9211` reads:

> `TILE_UP_DOWN_ODD` splits an odd valid-row count so AIV0 receives
> `ceil(valid_rows / 2)` rows and AIV1 receives `floor(valid_rows / 2)` rows.

This reads as if the split is derived for you. It is not. `popVecTileFromGMFiFo`
computes both the offset and the GM row stride from the *consumer's own*
`tile.GetValidRow()` / `GetValidCol()`:

```cpp
gmStrideR    = splitNum * tile.GetValidCol() + 2 * subBlockId - 1;          // LEFT_RIGHT_ODD
subAIVOffset = subBlockId * (tile.GetValidRow() + subBlockId) * ... ;       // UP_DOWN_ODD
```

Those identities only hold when AIV0 passes `ceil` and AIV1 passes `floor`, i.e. the
frontend must feed a `get_subblockid()`-dependent `valid_row` / `valid_col` operand
into `tpop`. Passing the full valid shape, or the same value on both sub-cores,
silently yields wrong addresses and strides. Worth one explicit sentence in the
manual and the design doc, since it is a new requirement relative to the even modes
where both halves are equal.

---

### Smaller notes

- `docs/designs/ptoas-tpush-tpop-design.md:459` now says the even modes derive
  offsets from the `GlobalData` shape while ODD uses the runtime valid shape. After
  `505dbfa6` the even GM modes were *also* rewritten to use `tile.GetValidRow()` /
  `GetValidCol()` rather than the static `Cons`/`Prod` constants, so that sentence is
  already stale for `TILE_UP_DOWN` / `TILE_LEFT_RIGHT` too.
- License headers on the two new/updated `.pto` tests are correct, and
  `Passes.td:470` was kept in sync with the pass diagnostic. Good on the cross-layer
  checklist.

lib/PTO/Transforms/PTOToEmitC.cpp
```diff
@@ -1114,6 +1114,10 @@ static FailureOr<std::string> getTileSplitToken(int64_t split) {
...
+  case 4:
+    return std::string("TileSplitAxis::TILE_LEFT_RIGHT_ODD");
```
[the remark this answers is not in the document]
zhangstevenunity: Confirming this one, with a correction to the mechanism: it is not that TALLOC *cannot* compute an offset, it is that it silently computes the wrong one.

`include/pto/npu/a5/TAlloc.hpp` (untouched by pto-isa `505dbfa6`, which only added ODD to `popVecTileFromGMFiFo`):

```cpp
template <typename GlobalData, TileSplitAxis Split>
PTO_INTERNAL uint64_t getSubAIVOffset()
{
    constexpr int prodN = GlobalData::staticShape[DIM_4];
    constexpr int prodM = GlobalData::staticShape[DIM_3];
    if constexpr (Split == TileSplitAxis::TILE_UP_DOWN) {
        return get_subblockid() * prodM * prodN * sizeof(RawDType);
    } else { // TILE_LEFT_RIGHT
        return get_subblockid() * prodN * sizeof(RawDType);
    }
    ...
}
```

The chain ends in an unguarded `else`, so `TILE_UP_DOWN_ODD` and `TILE_LEFT_RIGHT_ODD` both fall into the even left-right branch. For `TILE_UP_DOWN_ODD` that is the wrong axis entirely; for `TILE_LEFT_RIGHT_ODD` it is the even column offset computed from the static full-slot `prodN` instead of the per-sub-core ceil/floor column count. This applies on `is_v2c_gm` and `is_both_gm`; for `is_c2v_gm` TALLOC adds no sub-core offset, which is correct there because the cube producer writes the whole slot.

Same root cause as my P1-2 in the top-level review: `505dbfa6` taught only the consumer GM read path about ODD, so every producer-side helper (`TALLOC_IMPL::getSubAIVOffset`, `pushAcc2VecFiFo`, `pushVec2MatFiFo`, `pushVec2GMFiFo`) either falls into an even-split `else` or matches no branch at all and emits nothing.

reedhecre:
## A5 板测成功
- 触发方式：`merged`
- 源码提交：`5b943a13fd6d`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260728_101206_merged_pr1008.log`
- 结果 TSV：`/root/ptoas-board-monitor-a5/logs/20260728_101206_merged_pr1008.tsv`

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`5b943a13fd6d`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260727_191358_merged_pr1008.log`

### 失败用例
- `comm_p2p` (run, exit=1)

reedhecre:
## A3 板测失败详情：PR #1008

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260727_191358_merged_pr1008/npu_validation/CommSync/comm_p2p_binding_variants/main.cpp:122)
EZ9999[PID: 306088] 2026-07-27-19:44:30.209.450 (EZ9999):  The error from device(chipId:1, dieId:1), serial number is 476, there is an exception of aivec error, core id is 18, error code = 0x10, dump info: pc start: 0x124600000000, current: 0x124600000164, vec error info: 0xc9000000bd, mte error info: 0x6d0600009a, ifu error info: 0x3015685020000, ccu error info: 0x1ce600000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600009a, fixp_error1 info: 0x6d, fsmId:0, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=3, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_binding_variants_kernel, fault kernel info ext=comm_p2p_binding_variants_kernel, program id=0, hash=16541545288638093324.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-27 19:44:32] ERROR: testcase failed (exit 1): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260727_191358_merged_pr1008/npu_validation/CommSync/comm_p2p/main.cpp:122)
EZ9999[PID: 320201] 2026-07-27-19:44:52.313.431 (EZ9999):  The error from device(chipId:1, dieId:1), serial number is 477, there is an exception of aivec error, core id is 23, error code = 0x10, dump info: pc start: 0x124600000000, current: 0x124600000168, vec error info: 0xa10000ff90, mte error info: 0x6f0600006f, ifu error info: 0x3015685020000, ccu error info: 0x40a414030000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x600006f, fixp_error1 info: 0x6f, fsmId:0, tslot:3, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=3, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_kernel, fault kernel info ext=comm_p2p_kernel, program id=0, hash=6260346515940177617.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-07-27 19:44:54] ERROR: testcase failed (exit 1): comm_p2p
```

</details>

PR #1013, opened 2026-07-27
VMI pad-8 vs CCE ONEPT reduce store issue

learning-chip:
- Minimal cannsim reproducer for the pad-to-8 reduce store vs ASC `ONEPT_B32` (feature gap 03 in https://github.com/hw-native-sys/PTOAS/pull/997)
- Perf gap: CCE RVEC ~63, VMI ~281 on Ascend950 sim; desired MI path still fails bisheng.

See `docs/repro/store_pad8/README.md` to reproduce.

Tracking issue: https://github.com/hw-native-sys/PTOAS/issues/1022


mouliangyu:
I think we should first separate a kernel-writing issue from a compiler lowering issue here.

For each accumulator, the VMI semantics are already a `1 x T -> 1` reduction:

```python
red = pto.vmi.vcadd(acc_v, mask_t, reassoc=True)
pto.vmi.vstore(red, reduced, i, mask1)
```

`vcadd` already returns one logical value. Since the next consumer is only a store, broadcasting that value back to 8 lanes and writing a padded `[N_ACC, 8]` buffer is not semantically required:

```python
red = pto.vmi.vcadd(acc_v, mask_t, reassoc=True)
padded = pto.vmi.vbrc(red, size=8)
pto.vmi.vstore(padded, reduced_pad, i * 8, mask8)
```

I tested a self-contained direct `vcadd -> mask1 vstore` kernel with the current PTOAS build and CANN 9.0. It lowers through VPTO/LLVM IR and Bisheng produces a device object, without pad-8 staging.

The current lowering is still not ideal: the direct form becomes roughly

```text
vcadd
vdup 0
vadd PAT_VL1
predicate materialization
masked vsts
```

instead of the CCE `vcadd + ONEPT store` sequence. That may be a valid lowering optimization, but it is different from saying VMI cannot express or compile the compact result.

Could we add the following three-way measurement before attributing the whole gap to missing ONEPT support?

1. current pad-8 workaround;
2. direct `1 x T reduce -> mask1 store` with a compact `[N_ACC]` output;
3. CCE `vcadd + ONEPT` baseline.

This would tell us how much of the reported gap comes from the padded kernel formulation and how much remains in PTOAS lowering.

learning-chip:
> 2\. direct `1 x T reduce -> mask1 store` with a compact `[N_ACC]` output;

I already tried this in [this commit](https://github.com/learning-chip/vmi-demo/pull/26/changes/67f3fc3dc55afc4483a3ab7dbe4d091e8dfa7305#diff-c475f3a1a155132e1be34a06452972a5599eb065d128a02711417affb5765432), "VMI mask1" variant, still much slower than CCE ONEPT store...


Added a more focused commit https://github.com/hw-native-sys/PTOAS/pull/1013/changes/8f321ea5b72802af96cbf6f75f83825fc6a7dbd6 here, not much faster than pad8, still need ONEPT scalar store in ptoas vmi.

@mouliangyu

learning-chip:
group=1 works https://github.com/hw-native-sys/PTOAS/pull/1013/changes/c5345dff1019a35d594f4dc1e5e66ad87938ac37

PR #1017, opened 2026-07-27
Multi path fusion plan

Melikano:
## Summary

  Improve tile-fusion planning by exploring competing fusion groups instead of committing to the first acceptable DFG path.

  ## Key changes

  - Add whole-group cost evaluation alongside existing append legality checks.
  - Reuse shared cost computation for append and group evaluation.
  - Explore multiple candidate groups, retaining up to 64 candidates per group size.
  - Rank candidates by:
      1. Whole-group cost.
      2. Group size.
      3. Stable block/node order.

  - Preserve the existing append acceptance rules and fusion metadata contract.
  - Add regression coverage for:
      - Selecting a higher-value multi-step path.
      - Stable equal-cost tie-breaking.
      - Formation of subsequent independent fusion groups.
      
  ## Testing

  llvm-lit -sv build/test/lit \
    --filter 'fusion_plan_cost_guided_(search|tie)'

### DIFF
--- a/docs/designs/vf_costmodel_external_planner_protocol.md
+++ b/docs/designs/vf_costmodel_external_planner_protocol.md
@@ -1,12 +1,14 @@
-# PTOAS + VfSim Costmodel 接口设计
+# PTOAS + VfSim Cost Model Interface Design
 
-本文描述 PTOAS 以源码级 submodule 方式接入 VfSimulator costmodel 的当前接口形式。
-当前实现面向 A5 tile fusion 路径：PTOAS 负责生成合法 fusion group，VfSim 基于
-已选 group 做 costmodel 优化决策，并把结果写回同一份 MLIR IR。
+This document describes the current interface through which PTOAS integrates the
+VfSimulator cost model as a source-level submodule. The current implementation
+targets the A5 tile-fusion path: PTOAS generates legal fusion groups, VfSim uses
+the selected groups to make cost-model optimization decisions, and VfSim writes
+the results back to the same MLIR IR.
 
-## 接入形式
+## Integration Model
 
-VfSimulator 作为 PTOAS 的 git submodule 引入：
+VfSimulator is included in PTOAS as a git submodule:
 
 ```text
 PTOAS/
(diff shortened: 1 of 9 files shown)


Melikano:
# DeepSeek V4 Fusion Planner IR Comparison

This note summarizes what the current DeepSeek V4 smoke IR reports show about the new PTOAS fusion planner. The comparison is IR-only up to the PTOAS tile fusion planning and scheduling points. End-to-end VPTO/a5sim performance is not yet a valid comparison because several generated DeepSeek V4 kernels still hit A5 TileLib coverage gaps during later lowering.

## Executive Summary

The new fusion planner does improve fusion grouping on the DeepSeek V4 smoke IR set, but the benefit is localized in the currently compilable-to-fusion subset.

Across the smoke fusion IR reports:

| Result | Logical cases |
| --- | ---: |
| Old and new grouping differ | 4 |
| Old and new grouping match | 123 |
| Fusion IR dump failure | 40 |
| Total logical cases seen | 167 |

The four cases with differences are:

```text
decode_attention_csa/comb_sinkhorn
decode_attention_hca/comb_sinkhorn
decode_attention_swa/comb_sinkhorn
gate/gate
```

In the differing cases, the new planner produces fewer and larger fusion groups. The recurring pattern is that the legacy planner leaves small producer subgraphs as separate groups, while the new search planner merges them into a larger consumer group when the dependency structure allows it.

## Main Benefit Observed

The new pass finds cross-branch fusion opportunities that the old planner misses.

Instead of only forming local linear groups, the new planner can combine related producer branches into the main reduction or elementwise chain. This is visible in two important patterns:

1. A standalone elementwise branch is merged into a larger consumer chain.
2. Small `trowexpanddiv -> trowexpanddiv -> tadd` groups are merged into larger Sinkhorn normalization groups.

This matters because larger fusion groups usually give later lowering more opportunity to avoid intermediate materialization, reduce scheduling boundaries, and keep tile values closer to their consumers.

## Example 1: `gate/gate`

The legacy planner produces 3 groups after fusion planning:

```text
group 0:
  pto.trowexpandmul -> pto.tmaxs -> pto.tmax -> pto.tadd
  -> pto.tmins -> pto.tmins -> pto.texp -> pto.tmul -> pto.tmax

group 1:
  pto.texp -> pto.tadds

group 2:
  pto.tsubs -> pto.tmaxs
```

The new planner produces 2 groups:

```text
group 0:
  pto.trowexpandmul -> pto.tmaxs -> pto.tadd
  -> pto.tsubs -> pto.tmaxs
  -> pto.tmins -> pto.tmins -> pto.texp -> pto.tmul -> pto.tmax

group 1:
  pto.texp -> pto.tadds
```

The practical improvement is that the old standalone:

```text
pto.tsubs -> pto.tmaxs
```

branch is absorbed into the larger main fusion group by the new planner.

This difference survives the following `pto-op-scheduling` pass:

| Stage | Old planner | New planner |
| --- | ---: | ---: |
| After fusion plan | 3 groups | 2 groups |
| After scheduling | 3 groups | 2 groups |

So this is not just a temporary annotation difference before scheduling. The scheduled IR still exposes a larger group for the new planner.

## Example 2: `decode_attention_* / comb_sinkhorn`

All three attention variants show the same high-level pattern.

The legacy planner produces 11 groups. The new planner produces 8 groups.

The old planner repeatedly leaves a small group like this:

```text
pto.trowexpanddiv -> pto.trowexpanddiv -> pto.tadd
```

next to a larger normalization/division group like this:

```text
pto.trowexpanddiv -> pto.trowexpanddiv
-> pto.tadd -> pto.tadd -> pto.tadds
-> pto.tdiv -> pto.tdiv -> pto.tdiv -> pto.tdiv
```

The new planner merges those pieces into a larger 10-op group:

```text
pto.trowexpanddiv -> pto.trowexpanddiv
-> pto.tadd -> pto.tadd -> pto.tadd -> pto.tadds
-> pto.tdiv -> pto.tdiv -> pto.tdiv -> pto.tdiv
```

The group count difference is stable after scheduling:

| Case | Old after plan | New after plan | Old after scheduling | New after scheduling |
| --- | ---: | ---: | ---: | ---: |
| `decode_attention_csa/comb_sinkhorn` | 11 | 8 | 11 | 8 |
| `decode_attention_hca/comb_sinkhorn` | 11 | 8 | 11 | 8 |
| `decode_attention_swa/comb_sinkhorn` | 11 | 8 | 11 | 8 |

This is the strongest evidence in the current reports that the new search planner is doing useful global grouping work, not just renumbering fusion IDs.

## Cases With No Observed Difference

Most successful fusion IR dumps currently show identical grouping between the old and new planners:

```text
123 logical cases matched
```

This does not mean the new planner is worse or inactive. It means that, for those kernels and shapes, either:

1. The legacy planner already finds the same profitable groups.
2. The IR does not contain a larger legal grouping opportunity.
3. Later scheduling normalizes away any temporary grouping difference.

For those cases, performance should be expected to match unless later lowering treats equivalent fusion annotations differently.

## Current Limits of the Evidence

This report should be treated as fusion-plan evidence, not end-to-end performance evidence.

The VPTO/a5sim path is still blocked for parts of DeepSeek V4 by A5 TileLib coverage gaps, including missing or mismatched templates such as high-precision/tmp-form `pto.trsqrt`. Because of that, we cannot yet claim measured simulator speedup from these larger groups.

A valid performance conclusion still requires:

1. Successful old and new compilation through VPTO for the same cases.
2. Identical simulator platform and options, especially `a5sim`.
3. Identical PyPTO inputs, shapes, fixtures, and runtime configuration.
4. Correctness passing for both old and new generated programs.
5. Comparable timing or cycle metrics collected from the same simulator interface.

Until those conditions hold, the safe conclusion is:

> The new fusion planner improves fusion grouping quality in selected DeepSeek V4 kernels by producing fewer, larger groups. The observed benefit is most visible in `gate/gate` and `comb_sinkhorn`, but end-to-end performance impact is not yet measurable from the current VPTO/a5sim run because later lowering fails on unrelated template coverage issues.

## Recommended Next Step

Use these four differing kernels as the first targets once the VPTO TileLib blockers are fixed. They are the best candidates for measuring whether the new grouping translates into simulator-visible improvement.

Priority order:

1. `gate/gate`
2. `decode_attention_csa/comb_sinkhorn`
3. `decode_attention_hca/comb_sinkhorn`
4. `decode_attention_swa/comb_sinkhorn`

The `comb_sinkhorn` cases are especially useful because they show the same structural improvement across three DeepSeek attention variants.

Issue #950, opened 2026-07-17
perf: ptoas 0.50 >160x compile slowdown on csa_slots_build_valid_qk_plan kernel vs 0.48


wangqin1723-max:
## Summary

ptoas **0.50** takes **>120s** to assemble a specific PTO kernel (`.pto`) that ptoas **0.48** assembles in **0.72s** — a **>160x** deterministic slowdown. Same input, same flags, host-only (no device involved). Other kernels are unaffected on 0.50; this one is the outlier.

## Reproduction (host-only, deterministic)

Input `.pto` (77242 bytes, ASCII): https://gist.github.com/wangqin1723-max/0d4a69927ecca9e56308d6c517b532ae
(raw gist — download as `csa_slots_build_valid_qk_plan.pto`)

```bash
ptoas csa_slots_build_valid_qk_plan.pto -o out.cpp \
  --enable-insert-sync --pto-level=level3 --pto-arch a3
```

| ptoas | wall | user | output `.cpp` |
|-------|------|------|---------------|
| **0.48** | **0.72s** | 0.59s | 40584 B ✅ |
| **0.50** | **>120s** (did not finish within 2 min) | — | (none) ❌ |

Measured on the login host (no device, no load contention), same `.pto`, same flags — only the ptoas version differs.

## The kernel

`csa_slots_build_valid_qk_plan` from PyPTO-Lib's DeepSeek-V4 **CSA decode** attention (`models/deepseek/v4/decode_sparse_attn.py`). It's a single fused CORE_GROUP scope (PyPTO-Lib PR #734) combining compressed-slot masking + softmax-bias (build_valid) + the load-balance plan cursor — i.e. a **scalar/cursor-heavy scope with loop-carried `if`-phi** (`pl.range` cursor loops, `pl.read`/`pl.write` of a `qk_order`/`qk_wcur` cursor, `if qk_block_valid > 0` branching), compiled under `memory_planner=PYPTO` (`--pto-level=level3`).

## Impact

[...]

PR #1082, opened 2026-07-31
Fix tile arith.select aliases in autosync

FangRui0:
## Summary

- Restore tile `arith.select` alias propagation in both autosync engines so dependencies through selected tile buffers are not dropped.
- Record the regression source: the old `arith.select` handling lived in `PTOIRTranslator::RecursionIR` behind a `MemRefType` gate and was removed by `22d4f58` (`Remove memref compatibility from sync analysis`) instead of being re-gated for tile types.
- Close the planner/sync asymmetry: legacy and modern memplanners kept unioning select roots, but InsertSync and GraphSyncSolver were blind to tile-native select aliases.
- Deduplicate legacy alias meminfo propagation to avoid exponential growth through repeated cheap alias ops.
- Strengthen regression coverage so reads through a select require waits for both candidate producers, and writes through a select update both candidate aliases.

## Validation

- Remote A3 build: `cmake --build build-pr1082 -j16` passed.
- Remote related lit: `llvm-lit -v build-pr1082/test/lit --filter 'arith_select_tile_sync|multi_tile_dyn_slot_conservative_sync|ptr_roundtrip_scalar_sync|multi_tile_alias_slot_dyn_event_id'` passed, 4 tests.
- Remote full lit: `cmake --build build-pr1082 --target check-pto -j16` passed, 1532 passed / 1533 discovered / 1 unsupported / 0 failures.

### DIFF
--- a/lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp
+++ b/lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp
@@ -357,6 +357,10 @@ void PTOIRTranslator::RecursionIR(Region *region) {
     else if (auto bitcast = dyn_cast<pto::BitcastOp>(op)) {
       UpdateAliasBufferInfo(bitcast.getResult(), bitcast.getSrc());
     }
+    else if (auto select = dyn_cast<arith::SelectOp>(op)) {
+      UpdateAliasBufferInfo(select.getResult(), select.getTrueValue());
+      UpdateAliasBufferInfo(select.getResult(), select.getFalseValue());
+    }
 
     // --- Case C: 控制流 (SCF) ---
     else if (auto forOp = dyn_cast<scf::ForOp>(op)) {
--- a/lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp
+++ b/lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp
@@ -82,6 +82,18 @@ static pto::AddressSpace getTileAddressSpace(pto::TileBufType type) {
   return pto::AddressSpace::MAT;
 }
 
+static void
+appendUniqueMemInfo(SmallVectorImpl<std::unique_ptr<BaseMemInfo>> &infos,
+                    std::unique_ptr<BaseMemInfo> candidate) {
+  if (!candidate)
+    return;
+  if (llvm::any_of(infos, [&](const std::unique_ptr<BaseMemInfo> &info) {
+        return info && *info == *candidate;
+      }))
+    return;
+  infos.emplace_back(std::move(candidate));
+}
+
 } // namespace
 
 static bool getConstIndexValue(Value value, int64_t &out) {
@@ -829,17 +841,12 @@ void PTOIRTranslator::UpdateAliasBufferInfo(Value result, Value source) {
 
   auto &resultMemInfoVec = buffer2MemInfoMap_[result];
   for (auto &parentInfo : buffer2MemInfoMap_[source])
-    resultMemInfoVec.emplace_back(parentInfo->clone(result));
+    appendUniqueMemInfo(resultMemInfoVec, parentInfo->clone(result));
 }
 
 void PTOIRTranslator::UpdateConservativeAliasBufferInfo(Value result,
                                                         Value source) {
-  if (!result || !source) return;
-  if (!buffer2MemInfoMap_.contains(source)) return;
-
-  auto &resultMemInfoVec = buffer2MemInfoMap_[result];
-  for (auto &parentInfo : buffer2MemInfoMap_[source])
-    resultMemInfoVec.emplace_back(parentInfo->clone(result));
+  UpdateAliasBufferInfo(result, source);
 }
 
 LogicalResult PTOIRTranslator::UpdateIntToPtrOpMemInfo(pto::IntToPtrOp op) {
--- a/test/lit/pto/arith_select_tile_sync.pto
+++ b/test/lit/pto/arith_select_tile_sync.pto
@@ -13,10 +13,31 @@
 // RUN:   | FileCheck %s --check-prefix=GSS
 
 module attributes {pto.target_arch = "a3"} {
-  func.func @arith_select_tile_sync(
+  func.func @arith_select_tile_read_sync(
       %cond: i1,
       %src0: !pto.partition_tensor_view<16x16xf32>,
-      %src1: !pto.partition_tensor_view<16x16xf32>,
+      %dst: !pto.partition_tensor_view<16x16xf32>)
+      attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
+    %addr0 = arith.constant 0 : i64
+    %addr1 = arith.constant 1024 : i64
+    %addr2 = arith.constant 2048 : i64
+    %a = pto.alloc_tile addr = %addr0 : !pto.tile_buf<vec, 16x16xf32>
+    %seed = pto.alloc_tile addr = %addr1 : !pto.tile_buf<vec, 16x16xf32>
+    %b = pto.alloc_tile addr = %addr2 : !pto.tile_buf<vec, 16x16xf32>
+
+    pto.tload ins(%src0 : !pto.partition_tensor_view<16x16xf32>)
+              outs(%a : !pto.tile_buf<vec, 16x16xf32>)
+    pto.tabs ins(%seed : !pto.tile_buf<vec, 16x16xf32>)
+             outs(%b : !pto.tile_buf<vec, 16x16xf32>)
+    %selected = arith.select %cond, %a, %b : !pto.tile_buf<vec, 16x16xf32>
+    pto.tstore ins(%selected : !pto.tile_buf<vec, 16x16xf32>)
+               outs(%dst : !pto.partition_tensor_view<16x16xf32>)
+    return
+  }
+
+  func.func @arith_select_tile_write_sync(
+      %cond: i1,
+      %src: !pto.partition_tensor_view<16x16xf32>,
       %dst: !pto.partition_tensor_view<16x16xf32>)
       attributes {pto.kernel_kind = #pto.kernel_kind<vector>} {
     %addr0 = arith.constant 0 : i64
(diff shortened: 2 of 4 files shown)


test/lit/pto/arith_select_tile_sync.pto
```diff
@@ -0,0 +1,52 @@
...
+    pto.tstore ins(%selected : !pto.tile_buf<vec, 16x16xf32>)
+               outs(%dst : !pto.partition_tensor_view<16x16xf32>)
+    return
+  }
+}
+
+// INSERT-LABEL: func.func @arith_select_tile_sync
+// INSERT: pto.tload
+// INSERT: pto.tload
+// INSERT: pto.set_flag[<PIPE_MTE2>, <PIPE_MTE3>,
```
zhangstevenunity, line 88: **This test still passes when half of the fix is removed.**

Both candidates are written by `pto.tload`, so both producers sit on the same in-order MTE2 pipe. InsertSync's transitive elimination (`InsertSyncAnalysis.cpp:426` `isAlreadySync` -> `UpdateSyncRecord`, keyed on *pipe*, not on buffer) therefore collapses the two RAW dependencies down to the later producer alone -- and the later producer is the `false` operand. The GSS side collapses the same way.

I checked this rather than reasoning about it: patching `PTOIRTranslator.cpp` to keep only `getFalseValue()`, and `SyncSolverIRTranslator.cpp` to push only `select.getFalseValue()`, then rebuilding, leaves both the `INSERT` and `GSS` FileChecks passing on byte-identical output. So the test locks in "the select is traversed at all" (the pre-PR bug), but not "both candidate tile values are visited", which is what the summary claims and what the second `UpdateAliasBufferInfo` call exists for.

Dropping the *true* operand instead is caught, but only incidentally: the `set_flag` then moves up to before the second `pto.tload`, which trips the check on this line.

A discriminating shape needs the two candidates written on *different* pipes, so no per-pipe elimination applies:

```mlir
    pto.tload ins(%src0 : !pto.partition_tensor_view<16x16xf32>)
              outs(%a : !pto.tile_buf<vec, 16x16xf32>)          // MTE2
    pto.tabs ins(%c : !pto.tile_buf<vec, 16x16xf32>)
             outs(%b : !pto.tile_buf<vec, 16x16xf32>)           // V
    %selected = arith.select %cond, %a, %b : !pto.tile_buf<vec, 16x16xf32>
    pto.tstore ins(%selected : !pto.tile_buf<vec, 16x16xf32>)
               outs(%dst : !pto.partition_tensor_view<16x16xf32>)   // MTE3
```

Checks below are the verified output of this branch. Both engines emit both waits, but they order the two waits differently (InsertSync emits V first, GSS emits MTE2 first), so the pair needs `-DAG` if you share one check block:

```
// CHECK: pto.tload
// CHECK: pto.set_flag[<PIPE_MTE2>, <PIPE_MTE3>,
// CHECK: pto.tabs
// CHECK: pto.set_flag[<PIPE_V>, <PIPE_MTE3>,
// CHECK: arith.select
// CHECK-DAG: pto.wait_flag[<PIPE_MTE2>, <PIPE_MTE3>,
// CHECK-DAG: pto.wait_flag[<PIPE_V>, <PIPE_MTE3>,
// CHECK: pto.tstore
```

Under the `false`-only mutation this shape loses the `MTE2 -> MTE3` pair entirely and keeps only `V -> MTE3`, so it fails loudly. Keeping the current two-`tload` function as well is still useful, it just needs a companion that can actually fail.

lib/PTO/Transforms/InsertSync/PTOIRTranslator.cpp
```diff
@@ -357,6 +357,10 @@ void PTOIRTranslator::RecursionIR(Region *region) {
     else if (auto bitcast = dyn_cast<pto::BitcastOp>(op)) {
       UpdateAliasBufferInfo(bitcast.getResult(), bitcast.getSrc());
     }
+    else if (auto select = dyn_cast<arith::SelectOp>(op)) {
+      UpdateAliasBufferInfo(select.getResult(), select.getTrueValue());
+      UpdateAliasBufferInfo(select.getResult(), select.getFalseValue());
```
zhangstevenunity, line 374: `UpdateAliasBufferInfo` (`:826-833`) appends without dedup, so the merge rule established here is `|map[sel]| = |map[true]| + |map[false]|`. That is additive per edge but multiplicative over a DAG, and this PR makes such a DAG constructible out of one cheap op.

Measured on a build of this branch, paired ping-pong chain (`%a_i = arith.select %c_i, %a_{i-1}, %b_{i-1}` and `%b_i = arith.select %c_i, %b_{i-1}, %a_{i-1}`, distinct conditions so nothing folds), `--pto-arch=a3 --pto-level=level3`, wall time and peak RSS:

| stages | `--enable-insert-sync` | `--enable-graph-sync-solver` | no autosync |
|---|---|---|---|
| 16 | 0.62s / 212 MB | | |
| 18 | 0.94s / 283 MB | | |
| 20 | 1.54s / 578 MB | | |
| 22 | 4.14s / 1742 MB | 1.90s / 191 MB | 1.87s / 189 MB |

Exactly 4x memory per two stages, i.e. 2^n entries at roughly 370 bytes each. It is memory-bound rather than time-bound, so a couple more stages exhausts a 16 GB machine rather than merely getting slow.

The GraphSyncSolver half of this same PR is structurally immune: `tracebackMemVals` (`SyncSolverIRTranslator.cpp:107-149`) carries a `visited` DenseSet and collects into a `SetVector`, so it stays linear in distinct SSA values, which is why the middle column is flat. `PTOPlanMemoryModern` dedups as well (`unionRoots`). The legacy translator is the only one of the three that does not.

This is strictly pre-existing behaviour of `UpdateAliasBufferInfo` -- `UpdateYieldOpInfo` can already produce multi-entry values through `scf.if` -- so I would not block the merge on it. But `BaseMemInfo::operator==` already exists (`SyncCommon.h:122`), and skipping an append whose value is already present caps the vector at the number of distinct roots for about three lines of code. Given that issue #950 was this same blowup shape in a neighbouring pass, it seems worth doing while you are in here.

zhangstevenunity (commented):
Reviewed by building this branch (LLVM 21.1.8 / clang, a3 / level3) and running experiments against the built compiler. Full `check-pto`: **1532 passed of 1533 discovered (1 unsupported), 0 failures**.

**The fix is correct, and it is a genuine regression fix rather than speculative hardening.** Worth recording in the PR description: the `arith.select` case used to exist in `PTOIRTranslator::RecursionIR` and was deleted by 22d4f58d ("Remove memref compatibility from sync analysis"), because it was gated on `isa<MemRefType>` -- when memref support went away the whole branch went with it instead of being re-gated on tile types. Both memory planners kept modelling the pattern the entire time (`PTOPlanMemory.cpp:615-618`, and `PTOPlanMemoryModern.cpp:912-918` which unions the roots), so only the two sync engines were blind to it. That asymmetry is what this PR closes.

What I verified beyond the PR's own test, all on a build of this branch:

- **Conservative in both directions.** Reading through the select produces dependencies against both candidates; writing through it (`pto.tload ... outs(%selected)`) records the WRITE against both, so later reads of either candidate are ordered.
- **Distinct pipes get distinct flags.** With `%a` written by `pto.tload` (MTE2) and `%b` by `pto.tabs` (V), both engines emit both waits ahead of the `tstore`.
- **The realistic in-tree shape works.** `test/lit/pto/plan_memory_five_gates_phi_family_select.pto` (select inside `scf.if`, yielded out) run at `--pto-level=level2 --enable-insert-sync` now gets its MTE2 -> MTE3 pair around the `tstore`.
- **End to end, not just `--emit-pto-ir`.** Full compilation emits the select as a C++ ternary over `Tile` objects with the flags around it.

Two things worth addressing, neither blocking, details inline:

1. The regression test cannot fail for the property the PR summary claims to fix. I confirmed that by mutating the compiler and rebuilding, not by reading it.
2. `UpdateAliasBufferInfo` has no dedup, so the legacy half of this PR makes a 2^n meminfo blowup reachable through one cheap op. The GraphSyncSolver half of the same PR is structurally immune. Measurements inline.

Nit, no action needed for merge: `test/lit/pto/multi_tile_dyn_slot_conservative_sync.pto:11-18` still documents this exact mechanism ("The PTOIRTranslator now sees the arith.select on memref values and conservatively aliases both branches via `UpdateConservativeAliasBufferInfo`"). That has been describing deleted code since 22d4f58d, and it is doubly stale now that sync deliberately runs *before* `PTOResolveBufferSelect` (`tools/ptoas/ptoas.cpp:3398-3423`), so the dynamic-slot select chain is never seen by sync at all. This PR is a natural place to correct it.

zhangstevenunity (approved):
Re-reviewed at f8ce971 (previously 29eb583d). All three findings from my earlier review are addressed, and I verified each one locally rather than taking the diff at face value. Approving.

Built the PR head in-tree (LLVM21, a3) and ran the full suite: 1532 passed / 1533 discovered / 1 unsupported / 0 failures, matching your remote run.

## 1. The regression test now actually discriminates

My earlier objection was that the test could not fail: both select candidates were written by `pto.tload`, i.e. the same in-order MTE2 pipe, so InsertSync's per-pipe transitive elimination collapsed the two RAW deps into one -- aliasing only `getFalseValue()` left both FileChecks passing on byte-identical output.

The new `arith_select_tile_cross_pipe_read_sync` (tload/MTE2 vs tabs/V) fixes that. I re-ran the mutation experiments against the new test:

| Mutation (both engines unless noted) | Result |
|---|---|
| Drop `getFalseValue()` | FAIL -- caught at all 3 INSERT functions |
| Drop `getTrueValue()` | FAIL -- caught by the cross-pipe function |
| Drop `getFalseValue()` in GSS only | FAIL -- caught at all 3 GSS functions |

The third case is worth calling out: lit stops at the first failing RUN line, so when both engines are mutated the GSS checks never execute at all. Mutating GSS alone confirms that half of the test carries weight too. Both engines and both operands are now genuinely covered.

I also confirmed the emitted IR places both waits immediately before the tstore, in V-then-MTE2 order -- so the `CHECK-DAG` is load-bearing rather than decorative:

```
pto.set_flag[<PIPE_MTE2>, <PIPE_MTE3>, <EVENT_ID0>]
pto.tabs ins(%1 ...) outs(%2 ...)
pto.set_flag[<PIPE_V>, <PIPE_MTE3>, <EVENT_ID0>]
%3 = arith.select %arg0, %0, %2 : !pto.tile_buf<vec, 16x16xf32>
pto.wait_flag[<PIPE_V>, <PIPE_MTE3>, <EVENT_ID0>]
pto.wait_flag[<PIPE_MTE2>, <PIPE_MTE3>, <EVENT_ID0>]
pto.tstore ins(%3 ...)
```

## 2. The dedup fixes the exponential growth, and is sound

A/B on the paired ping-pong select chain (distinct non-foldable conditions, a3/level3, peak RSS):

| n | without dedup | with dedup |
|---|---|---|
| 16 | 210 MB | 188 MB |
| 18 | 283 MB | 190 MB |
| 20 | 578 MB | 189 MB |
| 22 | 1741 MB | 189 MB |

Reverting just the `appendUniqueMemInfo` call reproduces the 4x-per-2-stages curve exactly; with it, memory is flat at the ~189 MB no-autosync baseline.

On soundness, `appendUniqueMemInfo` can only drop genuinely identical entries: `BaseMemInfo::operator==` compares all seven fields, `BaseMemInfo` has no subclasses and no virtuals (so the non-virtual compare is complete, not a slicing hazard), and every downstream transform of a meminfo entry is a pure function of those fields -- `UpdateSlotSelectedAliasBufferInfo`'s const-slot narrowing and `UpdateTileSubViewAliasBufferInfo`'s offsetting both derive their result solely from the parent's fields. Equal entries therefore can never diverge later, so dropping one is safe. I also checked that no consumer reads only the first meminfo entry.

Those two helpers still `emplace_back` undeduped, but each is 1:1 (one source, same entry count), so they are additive rather than multiplicative -- the merge point that actually produced the blowup is exactly the one you deduped.

## 3. Stale comment

`multi_tile_dyn_slot_conservative_sync.pto` is corrected, and the replacement text checks out: the sync passes are added at `tools/ptoas/ptoas.cpp:3404-3418`, `PTOResolveBufferSelect` at :3423, so InsertSync genuinely never sees the lowering-time select chain.

## Other checks

- Placement is correct in both if-else chains: nothing earlier in `RecursionIR` matches `arith.select`, and in `tracebackMemValsStep` the new branch precedes the generic alias-info/DPS fallbacks, which would not have matched it anyway.
- The absence of a type gate on the select branch is fine -- `UpdateAliasBufferInfo` no-ops when the source has no meminfo, so a scalar select costs two map misses. A select over `ptrtoint` addresses now aliases both, which is the conservative and correct direction.
- `UpdateConservativeAliasBufferInfo` was byte-identical to `UpdateAliasBufferInfo` before this PR, so collapsing it to a delegation is behavior-preserving (it now additionally picks up the dedup).

Nothing blocking. Thanks for closing the loop on all three findings.

reedhecre:
## A5 板测成功
- 触发方式：`merged`
- 源码提交：`62669cf21dae`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260804_121305_merged_pr1082.log`
- 结果 TSV：`/root/ptoas-board-monitor-a5/logs/20260804_121305_merged_pr1082.tsv`

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`62669cf21dae`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260803_214454_merged_pr1082.log`

### 失败用例
- `comm_p2p` (run, exit=1)

reedhecre:
## A3 板测失败详情：PR #1082

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260803_214454_merged_pr1082/npu_validation/CommSync/comm_p2p_binding_variants/main.cpp:122)
EZ9999[PID: 2676436] 2026-08-03-22:19:04.130.825 (EZ9999):  The error from device(chipId:0, dieId:1), serial number is 2572, there is an exception of aivec error, core id is 26, error code = 0x10, dump info: pc start: 0x124200000000, current: 0x124200000164, vec error info: 0xdf00005f59, mte error info: 0xa5030000cd, ifu error info: 0x20002f60d79c0, ccu error info: 0x40e008000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x30000cd, fixp_error1 info: 0xa5, fsmId:1, tslot:6, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=1, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_binding_variants_kernel, fault kernel info ext=comm_p2p_binding_variants_kernel, program id=0, hash=3889294368825709812.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-08-03 22:19:06] ERROR: testcase failed (exit 1): comm_p2p_binding_variants
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260803_214454_merged_pr1082/npu_validation/CommSync/comm_p2p/main.cpp:122)
EZ9999[PID: 2688772] 2026-08-03-22:19:23.774.424 (EZ9999):  The error from device(chipId:0, dieId:1), serial number is 2573, there is an exception of aivec error, core id is 31, error code = 0x10, dump info: pc start: 0x124200000000, current: 0x124200000168, vec error info: 0x4000000b7, mte error info: 0x7503000008, ifu error info: 0x212c0d6200800, ccu error info: 0x40e008000000004f, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x10, 0, 0) errorStr: Illegal instruction, which is usually caused by unaligned UUB addresses. fixp_error0 info: 0x3000008, fixp_error1 info: 0x75, fsmId:0, tslot:6, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
       [DFX_INFO]Aicore kernel execute failed, device_id=1, stream_id=46, report_stream_id=46, task_id=0, flip_num=0, fault kernel_name=comm_p2p_kernel, fault kernel info ext=comm_p2p_kernel, program id=0, hash=15664819858989799263.[FUNC:GetError][FILE:stream.cc][LINE:1475]
[2026-08-03 22:19:25] ERROR: testcase failed (exit 1): comm_p2p
```

</details>

PR #1087, opened 2026-07-31
fix(vmi): VMI legality checks, vsstb op, and PTODSL VMI surface enforcement

### LINKED ISSUE #1058: [Feature] Validate VMI register types in PTODSL and TileLang frontends
### Summary

Add frontend validation for VMI register types in both PTODSL and TileLang. Invalid VMI register element types should be rejected before generating or lowering PTO/VMI IR, instead of failing later in an MLIR verifier, lowering, code generation, or compilation.

### Motivation / use case

PTODSL and TileLang users can currently construct VMI programs with register types that are not legal for VMI. The error is discovered too late, so diagnostics are farther from the source declaration and may surface as a confusing backend failure.

Both frontends should provide the same early and actionable validation behavior. This also prevents PTODSL and TileLang from accepting different type sets for equivalent VMI programs.

### Proposed API / behavior

- Define or reuse a single source of truth for the VMI register element types supported by PTOAS, including integer signedness/width, floating-point and low-precision types as applicable.
- Validate VMI register types when PTODSL and TileLang construct or compile VMI values/operations, before IR lowering or code generation.
- Reject an unsupported type with a frontend diagnostic that identifies the offending register/value, the supplied type, and the expected legal type set.
- Apply the validation consistently to explicit register declarations and inferred/result register types.
- Keep the existing IR/backend verifier checks as a defensive fallback.
- Do not change the behavior of programs that use legal VMI register types.

### Acceptance criteria

- PTODSL rejects an illegal VMI register type at frontend compile time with an actionable diagnostic.
- TileLang rejects the equivalent illegal type at frontend compile time with an equivalent diagnostic.
- Focused negative tests cover illegal explicit and inferred/result types in both frontends.
- Positive tests confirm representative legal integer, floating-point, and supported low-precision VMI register types remain accepted.
- The legal type definition is shared or otherwise kept centralized so the two frontends cannot silently drift.

### Alternatives considered

Relying only on the existing VMI IR verifier or downstream compiler errors preserves correctness, but reports the problem too late and produces a worse frontend user experience.

Zhendong404:
## Summary

Strengthens VMI legality validation and closes PTODSL surface gaps, rebased on current `hwsys/main`.
Fixes #1058 

### New op
- **`feat(vmi): add vmi.vsstb implementation`** — adds the block-strided store op with a required mask and fixed zero repeat stride, plus lowering guards: `pmode="merge"` is now rejected by both the verifier (clear diagnostic) and the legacy lowering instead of being silently dropped.

### VMI legality checks
- **`fix(vmi): improve legality check for vmi.vload/vstore`** — memory element compatibility, block-stride/mask rules, and group divisibility are checked in all mode branches.
- **`fix(vmi): enhance vcvt conversion attributes`** — rounding/saturation attribute validation with positive/negative tests.
- **`fix(vmi): enforce legal lane counts for vreg and mask`** — unified group verifiers align with the documented domain.
- **`fix(vmi): add UB-backed memory verification for gather/scatter`** — `vgather`/`vgatherb`/`vscatter` reject non-UB memory operands at verify time.

### PTODSL surface
- **`fix(ptodsl): enforce VMI lane-count whitelist on vinterpret_cast`** — the lane count derived by `vinterpret_cast` is checked against `{1,2,4,8,64,128,256}` (previously `8xf32 -> si8` silently produced `32` lanes, bypassing the constructor whitelist); the source layout is only preserved when the lane count is unchanged, and a layout-assigned source across a lane-count change is rejected instead of reusing a stale group/slot layout.
- **`test(ptodsl): improve coverage of vmi dsl tests`** — inventory/emission coverage for the formal `pto.vmi` namespace.

### Test fixes
- **`fix(test): fix failed lit testcases`** — updates `tadd_*` and `soft_postupdate` expectations to the deterministic lowering output (constant ordering, `set_mask_norm`, shared vs duplicated `-1` constants) and fixes the `vmi_gather_scatter_address_space_invalid` RUN line.

## Validation
- `check-pto`: 1538/1539 pass (1 unsupported: vfsim-costmodel feature off)
- `check-dsl`: 25/25 pass

### Root cause

`pto.merge_jit_modules()` built each kernel in an independent MLIR `Context`, then directly cloned operations from later modules into the first module. The cloned operations retained types and attributes owned by the source context. When that context was destroyed, destruction of the merged module accessed freed MLIR objects and could terminate with `SIGSEGV` in `mlir::Operation::~Operation()`.

### Fix

Reparse each source module text in the first module's target `Context` before cloning its operations. This keeps all operations, types, attributes, and dialect objects in the merged module under one Context lifetime.

### Validation

- `ptodsl/tests/test_ast_rewrite_example_ir.py`: PASS
- `check-dsl`: 29/29 passed
- `check-pto`: 1622 tests passed, 0 failures
- Valgrind: no invalid read/use-after-free or `pure virtual method called` on the merged AST softmax case

### DIFF
--- a/docs/isa/vmi-isa/00-architecture-overview.md
+++ b/docs/isa/vmi-isa/00-architecture-overview.md
@@ -61,11 +61,16 @@ E_v = 32 / sizeof(T) : lanes per VLane     (f32 → 8, f16/bf16 → 16, i8 → 3
 
 Logical vector register. `L` is the logical lane count; `T` is the element type.
 
-| T | bits | E_v (lanes per physical vreg) | Legal L multiples |
+| T | bits | E_v (lanes per physical vreg) | Legal L |
 |---|---|---|---|
-| `f32` / `i32` / `ui32` / `si32` | 32 | 64 | 64 |
-| `f16` / `bf16` / `i16` / `ui16` / `si16` | 16 | 128 | 64 |
-| `i8` / `ui8` / `si8` / `fp8_e4m3` / `fp8_e5m2` | 8 | 256 | 64 |
+| `f32` / `i32` / `ui32` / `si32` | 32 | 64 | `1, 2, 4, 8, 64, 128, 256` |
+| `f16` / `bf16` / `i16` / `ui16` / `si16` | 16 | 128 | `1, 2, 4, 8, 64, 128, 256` |
+| `i8` / `ui8` / `si8` / `fp8_e4m3` / `fp8_e5m2` | 8 | 256 | `1, 2, 4, 8, 64, 128, 256` |
+
+These are the legal lane counts on the formal public VMI/PTODSL surface.
+PTOAS currently accepts additional positive lane counts in internal VMI IR for
+lowering intermediates and compatibility tests; those are not public PTODSL
+type-construction choices.
 
 - **Full vector**: `L · bitwidth(T) == N · 2048` (integer multiple of 256B).
 - **Compact/partial vector**: `L · bitwidth(T) < 2048` — still backed by one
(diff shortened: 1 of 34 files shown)


reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`9c49c3697de3`
- 结果汇总：OK `220` / FAIL `0` / SKIP `3`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260805_012110_merged_pr1087.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260805_012110_merged_pr1087.tsv`

PR #1098, opened 2026-08-02
[PTODSL TileLib] add elemntwise 1d 2d versions

### LINKED ISSUE #1051: [VPTO][PTODSL TileLib] Support 1D/2D template selection for all element-wise TileOps
## Background

The VPTO backend currently provides PTODSL TileLib implementations for the
element-wise TileOps. Most implementations traverse a tile as a two-dimensional
region: an outer loop walks through rows and an inner vector loop processes the
columns of each row.

For a row-major tile whose valid region is physically contiguous, the same
operation can instead flatten the valid region and process it with a single
vector loop. This avoids per-row loop control and repeated row-tail handling.

The A5 implementation in [pto-isa](https://gitcode.com/cann/pto-isa.git)
already follows this model. For example,
[`TBinOp.hpp`](https://gitcode.com/cann/pto-isa/blob/master/include/pto/npu/a5/TBinOp.hpp)
contains separate 1D and 2D implementations and selects the applicable path
according to the tile shape and continuity conditions.

VPTO/PTODSL TileLib needs the same general capability. This issue is not only
about adding a 1D implementation for one binary operation. The goal is to make
**1D/2D template selection a common capability of all element-wise TileOps**.

## Scope

The scope covers all element-wise TileOps that currently have PTODSL TileLib
implementations for A5.

### Unary operations

- `tabs`
- `texp`
- `tlog`
- `tneg`
- `tnot`
- `trecip`
- `trelu`
- `trsqrt`
- `tsqrt`

### Tile-Tile operations

- `tadd`
- `tand`
- `tdiv`
- `tfmod`
- `tmax`
- `tmin`
- `tmul`
- `tor`
- `tprelu`
- `trem`
- `tshl`
- `tshr`
- `tsub`
- `txor`

### Tile-Scalar operations

- `tadds`
- `tands`
- `tdivs`
- `tfmods`
- `tlrelu`
- `tmaxs`
- `tmins`
- `tmuls`
- `tors`
- `trems`
- `tshls`
- `tshrs`
- `tsubs`
- `txors`

### Compare, select, conversion, and scalar fill operations

- `tcmp`
- `tcmps`
- `tsel`
- `tsels`
- `tcvt`
- `texpands`

The implementation may be delivered by operation family rather than in one
large change, but the common selection mechanism and the final acceptance must
cover the full list above. If an operation cannot legally use a flattened 1D
form because of its instruction semantics or data representation, that
exception must be documented and verified instead of silently omitted.

Reduction, row/column expansion, partial-update, load/store, data-movement,
sorting, random-number generation, and matrix operations are not part of this
issue. Although some of their implementations contain element-wise loops,
their TileOp semantics and layout requirements are different.

### 1. Provide reusable 1D and 2D element-wise template forms

PTODSL TileLib should provide reusable traversal forms for the major
element-wise families instead of implementing selection independently in every
operation:

- unary;
- Tile-Tile;
- Tile-Scalar;
- compare;
- select;
- conversion;
- scalar fill.

The 1D form flattens a legal valid region into one continuous element range and
uses one vector loop. The existing row-wise form remains the general 2D
fallback. Operation-specific computation, dtype restrictions, temporary
operands, masks, conversion modes, and scalar operand order must retain their
current semantics.

The template metadata must distinguish the two forms. In particular, the 1D
form should expose a one-dimensional loop structure and the 2D form should
retain its two-dimensional loop structure.

### 2. Define a common and conservative 1D legality rule

A 1D candidate may be selected only when every tile operand participating in
the traversal describes the same logical element range and that range can be
safely accessed as contiguous memory.

At a high level, the rule must consider:

- physical shape and valid shape;
- memory space;
- block layout and sub-layout;
- row stride or equivalent physical-continuity information;
- the shape and continuity of every input, output, mask, and temporary tile
  used by the operation;
- operation-specific restrictions such as predicate representation or
  source/destination dtype width changes.

For ordinary row-major tiles, a multi-row region is generally flattenable when
the valid column axis is full for every participating tile and there is no
padding or stride gap between adjacent rows. A single-row region may also be
flattenable. This should be encoded as an explicit shared constraint rather
than inferred only from an operation name or layout label.

Unknown or insufficient layout information must fall back to 2D. The 1D path
must never be selected speculatively.

Some element-wise operations need more specific rules:

- `tcmp`, `tcmps`, `tsel`, and `tsels` must account for predicate-mask layout
  and representation;
- `tcvt` must account for source/destination dtype widths and any packing or
  multi-step conversion form;
- operations carrying temporary Tile operands must include those operands in
  the legality decision;
- scalar operations must preserve normal and reverse operand forms where both
  are supported.

### 3. Make VPTO candidate selection deterministic

When both forms are legal, VPTO must select the 1D candidate. When the 1D
conditions are not met, VPTO must select the 2D candidate without changing the
TileOp result.

The complete selection path must preserve this decision:

```text
TileOp operand metadata
        -> PTODSL legality and ranking
        -> compact candidates attribute
        -> ExpandTileOp
        -> selected generated helper
```

Today the Python registry can rank candidates by priority, while the compiler
side stores and consumes a compact candidate list. The implementation must
ensure that priority/ordering is not lost between these stages and that the
expanded helper is the candidate selected by the legality and ranking rules.
Selection must not depend on Python import order, registration order, or an
accidental candidate ID order.

### 4. Keep specialization safe

All metadata that can change 1D legality or the generated helper body must be
available during candidate filtering and must participate in helper
specialization where applicable. Different shapes, valid shapes, layouts,
strides, memory spaces, operand forms, and conversion modes must not
incorrectly reuse the same generated helper.

### 5. Add family-level and end-to-end tests

Tests should demonstrate the common mechanism at the family level and ensure
that every operation in scope is registered with the expected 1D/2D behavior.
They do not need to duplicate every shape for every operation.

The test matrix should include, at minimum:

- contiguous multi-row full-axis input selecting 1D;
- contiguous single-row input selecting 1D;
- a multi-row tile with a non-full valid column axis selecting 2D;
- a stride gap, padding, or unsupported layout selecting 2D;
- vector-aligned and tail shapes;
- representative 8-bit, 16-bit, and 32-bit types supported by the operation
  families;
- representative unary, Tile-Tile, Tile-Scalar, compare, select, conversion,
  and scalar-fill operations;
- a catalog-level check proving that every operation listed in this issue has
  the required candidate coverage or an explicitly documented exception;
- VPTO expansion checks proving that the selected helper has the expected loop
  dimensionality;
- functional comparison showing identical results between legal 1D and 2D
  executions.

## Acceptance Criteria

The issue is complete when:

- all element-wise TileOps listed in the scope support deterministic 1D/2D
  selection, or have a reviewed and tested reason why a 1D form is not legal;
- eligible contiguous full-axis and single-row cases select a 1D helper;
- ineligible cases reliably retain the 2D helper;
- the 1D helper uses one flattened vector traversal rather than a row loop plus
  a column loop;
- tail processing is correct and neither path reads or writes outside the
  logical valid region;
- candidate selection remains consistent from PTODSL legality analysis through
  `ExpandTileOp`;
- selection is independent of template registration/import order;
- 1D and 2D executions produce equivalent results for the supported dtype and
  shape combinations;
- existing VPTO element-wise tests continue to pass;
- focused metadata-selection, expansion, functional, and performance tests are
  added;
- the shared legality and selection rules are documented for future
  element-wise TileOps.

## Performance Goal

Performance is evaluated by comparing the new 1D candidate with the existing
2D candidate for the **same TileOp, dtype, logical shape, data, compiler
options, and execution environment**. Comparison with another repository's
implementation is not required.

The expected direction is:

- the 1D path removes the outer row loop and its loop-control overhead;
- it avoids restarting vector traversal and tail state for every row;
- it does not add layout conversion, temporary-buffer movement, or extra
  vector computation;
- its vector compute instruction count is consistent with processing the
  flattened valid element count;
- it provides a measurable benefit for contiguous multi-row shapes where the
  row dimension creates meaningful 2D loop overhead;
- it is not slower than the 2D candidate for the selected benchmark set.

The performance report should cover representative operations from each major
family rather than only one operation. It should include multiple dtypes,
single-row and multi-row shapes, vector-aligned and tail shapes, and both 1D
eligible and 2D fallback layouts.

For each case, report at least:

- operation, dtype, physical shape, and valid shape;
- selected candidate and loop dimensionality;
- 1D and 2D execution cycles or latency under the same environment;
- relative improvement of 1D over 2D;
- confirmation that functional results are identical.

The optimization should demonstrate a stable overall benefit on 1D-eligible
element-wise workloads, with no material regression in cases that continue to
use the 2D fallback. Exact per-operation thresholds may be determined from the
benchmark baseline because different element-wise instructions have different
compute costs.

## Relevant Code Areas

- `ptodsl/ptodsl/tilelib/templates/a5/_elementwise.py`
- `ptodsl/ptodsl/tilelib/templates/a5/_remainder.py`
- the operation-specific A5 templates under
  `ptodsl/ptodsl/tilelib/templates/a5/`
- `ptodsl/ptodsl/tilelib/constraints.py`
- `ptodsl/ptodsl/tilelib/registry.py`
- `ptodsl/ptodsl/tilelib/serving/daemon.py`
- `lib/PTO/Transforms/InsertTemplateAttributes.cpp`
- `lib/PTO/Transforms/ExpandTileOp.cpp`
- PTODSL TileLib selection/render tests and VPTO lit tests

Melikano:
closes https://github.com/hw-native-sys/PTOAS/issues/1051


test/vpto/cases/elementwise-1d-2d-equivalence.py
```diff
@@ -0,0 +1,768 @@
...
+    _load_rows_to_ub(
+        src_ptr,
+        pto.f32,
+        4096,
+        rows,
+        cols,
+        TABS_PADDED_SHAPE[1],
+        4,
+    )
+    _wait_for_loads()
```
Zhendong404: Just use pto.tile.load and pto.tile.store in this st testcases. You can also use auto mode for an automatic insert sync.

test/vpto/cases/elementwise-1d-2d-equivalence.py
```diff
@@ -0,0 +1,768 @@
...
+    pto.wait_flag("MTE2", "V", event_id=0)
+
+
+def _wait_for_stores():
+    pto.set_flag("V", "MTE3", event_id=0)
+    pto.wait_flag("V", "MTE3", event_id=0)
+
+
+@pto.jit(
+    name="elementwise_tabs_1d_2d_equivalence",
```
Zhendong404: Maybe you can use the factory design pattern to simplify these testcases.

Zhendong404 (changes requested):

ptodsl/ptodsl/tilelib/serving/daemon.py
```diff
@@ -126,7 +126,9 @@ def _build_tile_specs(descriptor, operand_specs: list) -> dict:
             valid_shape=tuple(valid_shape) if valid_shape is not None else None,
             b_layout=config.get("b_layout", "row_major"),
             s_layout=config.get("s_layout", "none_box"),
+            s_fractal_size=config.get("s_fractal_size", 512),
```
Zhendong404: Regression risk for tiles without an explicit config attr. Both C++ producers leave `s_fractal_size` as `0` when `tileType.getConfigAttr()` is null or has no fractal attr (`InsertTemplateAttributes.cpp`: `int64_t fractalSize = 0;`; `ExpandTileOp.cpp` same pattern), and that 0 is sent in the JSON. Previously `TileSpec.mlir_type` hardcoded `fractal_size=512`, matching `TileBufConfigAttr::getDefault()`; now the 0 flows straight into `tile_buf_type`. The verifier in `lib/PTO/IR/PTOAttrs.cpp` (`TileBufConfigAttr::verify`) rejects anything outside {Mx, AB, C sizes} with "unsupported s_fractal_size", so a config-less tile that used to render fractal=512 now renders fractal=0 and either fails verification or mismatches the call-site type. Please normalize 0 -> 512 here (or emit the default on the C++ side) before consuming this field.

lib/TileOps/a5/_fillpad.py
```diff
@@ -52,8 +52,55 @@ def _zero(dtype):
...
+        static_valid_shape = getattr(tile, "static_valid_shape", None)
+    if static_valid_shape is None:
+        return None
+    dim = static_valid_shape[index]
+    if not isinstance(dim, int) or dim <= 0:
+        return None
+    return dim
+
+
+def _uses_custom_neg_one(op, src, dst):
```
Zhendong404: This looks unrelated to the 1D/2D selection feature and silently changes existing `tfillpad` semantics: any destination with a genuine `Zero` pad whose source valid cols are narrower than the destination valid cols now gets filled with -1 instead of 0. The comment itself says Zero is (ab)used as a placeholder for Custom(-1), i.e. real zero-padding users can no longer get zeros in this case. It also relies on probing private attributes (`_template_static_valid_shape` / `static_valid_shape`) that may not exist, in which case the behavior quietly reverts. This is not mentioned in the PR description or issue #1051 — please split it into its own PR with a documented compatibility decision.

lib/TileOps/a5/tcmps.py
```diff
@@ -5,97 +5,226 @@
...
             remained = valid_cols
             for col in range(0, valid_cols, lanes):
                 mask, remained = pto.make_mask(dtype, remained)
-                vec = pto.vlds(src[row, col:])
-                cmp = pto.vcmps(vec, scalar, mask, cmp_mode)
-                store_offset = (row * iters_per_row + col // lanes) * bytes_per_iter
-                pto.psts(cmp, dst_ptr, store_offset, dist=pto.PredicateDist.PK)
+                value = pto.vlds(src[row, col:])
+                cmp = pto.vcmps(value, scalar, mask, cmp_mode)
+                store_offset = row * dst_stride + col // 8
```
Zhendong404: This changes the 2D fallback's predicate layout, not just refactors it. The old `tcmps` template stored predicates flat-packed across rows (`(row * iters_per_row + col // lanes) * bytes_per_iter` over `total_elm = valid_rows * valid_cols`); the new 2D code writes at the physical row stride (`row * dst_stride + col // 8`). Whenever the destination's physical row bytes differ from the packed row bytes (padded predicate tile), old and new 2D produce different memory layouts for the same op. If this is a deliberate alignment with `tcmp`'s row-strided layout (with flat packing moved to the constrained 1D candidate), please state the dst layout contract for `tcmps` explicitly — otherwise existing consumers of the packed layout break silently.

ptodsl/ptodsl/tilelib/constraints.py
```diff
@@ -304,13 +328,455 @@ def require_contiguous(required=True):
...
+            return all(
+                shape[1] >= required_data_row_elements
+                for shape in data_shapes
+            )
+
+        data_rows_are_contiguous = all(
+            valid_shape[1] == shape[1]
+            for shape, valid_shape in zip(data_shapes, data_valid_shapes)
+        )
+        destination_holds_logical_range = predicate_row_bytes >= max(
```
Zhendong404: Unit mismatch: `predicate_row_bytes` is a byte count (`predicate_shape[1] * predicate_bytewidth`) while `max(shape[1] ...)` is a number of data *elements* per row. For e.g. f32 data with 256 columns this demands >= 256 predicate bytes/row where the actual packed need is ~32 bytes, so the check is only accidentally conservative and its intent is unclear. Since this gates whether `tcmps`-style flattened candidates take the flat-packed branch or the exact-stride branch, please either express it in terms of `required_row_bytes` or document what invariant it is meant to enforce.

Zhendong404 (commented):
Review of the 1D/2D element-wise selection infrastructure. Overall the constraint predicates, deterministic ordering and the new equivalence tests look solid; a few concrete issues and risks inline.

ptodsl/ptodsl/tilelib/constraints.py
```diff
@@ -353,6 +819,8 @@ def passes(predicates, context: dict) -> bool:
     "evaluate_candidate",
     "passes",
     "require_contiguous",
+    "require_conversion_1d",
+    "require_elementwise_1d",
```
Zhendong404, line 785: `require_predicate_compare_1d` and `require_predicate_select_1d` are exported from `ptodsl.tilelib.__init__` but missing from this module's own `__all__` (only `require_conversion_1d` / `require_elementwise_1d` were added). Explicit imports work, but the module's public surface is now inconsistent — please add both here too.

ptodsl/tests/test_tilelib_catalog.py
```diff
@@ -10,24 +10,25 @@
 import ast
 from pathlib import Path
 import unittest
+from pathlib import Path
```
Zhendong404: Duplicate import — `from pathlib import Path` already exists at the top of this file (line ~11).

lib/PTO/Transforms/InsertTemplateAttributes.cpp
```diff
@@ -889,7 +894,7 @@ parseCandidateAttributes(Operation *operation, StringRef metadataJson) {
   }
 
   auto *root = parsed->getAsObject();
-  auto *candidates = root ? root->getObject("candidates") : nullptr;
+  auto *candidates = root ? root->getArray("candidates") : nullptr;
```
Zhendong404, line 776: Removing the C++-side sort means candidate order — and therefore the selected template, since `ExpandTileOp` unconditionally takes `candidates[0]` — now depends entirely on the daemon returning a priority-sorted list. That holds for the in-tree daemon (`candidate_sort_key`), but any out-of-tree or older metadata helper that returns unsorted candidates will now silently select a different template instead of being normalized here. Consider either keeping a defensive sort, or documenting that the daemon/compiler protocol version must match. Related: `metadata_request` on the Python side now raises `AmbiguousTemplate` on a top-priority tie, turning a metadata query into a hard compile failure for any future op that registers two same-priority legal candidates — a lint/catalog-time check would be a friendlier place to catch that.

lib/TileOps/a5/tcvt.py
```diff
@@ -131,32 +228,103 @@ def _register_tcvt(
...
+    fallback = register_form(
+        traversal="2d",
+        candidate_name=name,
+        candidate_id=idx,
+        priority=0,
+    )
+    register_form(
+        traversal="1d",
+        candidate_name=f"{name}_1d",
+        candidate_id=38 + idx,
```
Zhendong404: The `38 + idx` magic offset (used again in `_register_tcvt_1d`) is fragile: it works today only because every 2D idx stays below 38 and the bespoke 1D registrations happen to pick non-colliding idx values. A future conversion with idx >= 38, or a missed offset, produces a duplicate candidate id and a compile-time "candidate ids must be unique" error far from the cause. Please allocate 1D candidate ids programmatically (e.g. a module-level counter or max(existing)+1) instead of a hand-maintained offset.

lib/TileOps/a5/tadd.py
```diff
@@ -17,9 +17,23 @@ def _vadd(lhs, rhs, mask):
...
+)
+
+
+template_tadd_1d = register_binary(
+    op="pto.tadd",
+    name="template_tadd_1d",
+    vector_op=_vadd,
+    dtypes=_DTYPES,
+    traversal="1d",
+    priority=10,
```
Zhendong404: `priority=10, candidate_id=1` is copy-pasted across ~40 op files as a magic convention. If a future candidate needs to sit between "default 2D" and "preferred 1D", every one of these files has to be renumbered by hand. A named constant (e.g. `PREFERRED_TRAVERSAL_PRIORITY`) in `_elementwise.py`, and ideally automatic candidate-id assignment in the register_* helpers, would make the convention explicit and less error-prone.

ptodsl/ptodsl/tilelib/constraints.py
```diff
@@ -304,13 +328,455 @@ def require_contiguous(required=True):
...
+
+def _is_static_rank2_shape(shape) -> bool:
+    return (
+        isinstance(shape, tuple)
+        and len(shape) == 2
+        and all(isinstance(dim, int) and dim > 0 for dim in shape)
+    )
+
+
+def _has_gap_free_row_stride(compact_mode) -> bool:
```
Zhendong404, line 733: Two maintainability notes on this area: (1) the kind/rank-2/valid<=physical/memory-space/layout/compact check sequence is now implemented four times — `_is_flat_local_tile` plus near-identical inline expansions in `require_elementwise_1d` and `require_conversion_1d`; please route all of them through the helper. (2) the accepted compact-mode literals (`{0, 1, "null", "normal", "Null", "Normal"}`) duplicate the alias table in `_types.py::_compact_mode_token` (which also knows `row_plus_one`); a single shared normalization helper would keep the two from drifting apart.

test/vpto/cases/elementwise-1d-2d-equivalence.py
Melikano: fixed: used pto.tile.load/store instead of mte manual transfers and used auto insert sync.

test/vpto/cases/elementwise-1d-2d-equivalence.py
Melikano: fixed: added a focused factory for structurally identical kernels while keeping specialized cases explicit.

ptodsl/ptodsl/tilelib/serving/daemon.py
Melikano: fixed: normalized a 0 fractal size to the default 512.

lib/TileOps/a5/_fillpad.py
Melikano: fixed: removed the unrelated Zero to Custom(-1) behavior and restored zero-padding semantics.

lib/TileOps/a5/tcmps.py
Melikano: fixed: restored the existing flat-packed predicate layout for the 2D compare fallback.

ptodsl/ptodsl/tilelib/constraints.py
Melikano: fixed: corrected the predicate-capacity check to compare byte counts consistently.

ptodsl/ptodsl/tilelib/constraints.py
Melikano: fixed: added the predicate compare and select constraints to constraints.__all__.

ptodsl/tests/test_tilelib_catalog.py
Melikano: fixed: removed the duplicate Path import.

lib/PTO/Transforms/InsertTemplateAttributes.cpp
Melikano: fixed: transported priority metadata, defensively sorted candidates in c++, and added explicit tie diagnostics.

lib/TileOps/a5/tcvt.py
Melikano: fixed: replaced the hard-coded tcvt ID offset with dynamic candidate-id allocation.

lib/TileOps/a5/tadd.py
Melikano: fixed: centralized traversal priorities and automatic candidate-id assignment in the shared helpers.

ptodsl/ptodsl/tilelib/constraints.py
Melikano: fixed: routed flat-tile checks through the shared helper and reused common compact-mode normalization.

Zhendong404 (approved):

reedhecre:
## A5 板测失败
- 触发方式：`merged`
- 源码提交：`3d798563b91e`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260811_170206_merged_pr1098.log`

### 失败用例

reedhecre:
## A3 板测失败
- 触发方式：`merged`
- 源码提交：`3d798563b91e`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260811_025526_merged_pr1098.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260811_025526_merged_pr1098.tsv`

### 失败用例
- `ci` (run, exit=2)

reedhecre:
## A3 板测失败详情：PR #1098

`stage=run` `info=exit=2`

```text
[ERROR] compare failed
[2026-08-11 03:18:49] ERROR: testcase failed (exit 2): ci
```

</details>

`stage=run` `info=exit=1`

```text
[ERROR] aclrtSynchronizeStream(stream) failed: 507035 (/home/zhongxuan/ptoas-board-monitor/runtime/runs/20260811_025526_merged_pr1098/npu_validation/Gather/gather_legacy/main.cpp:114)
EZ9999[PID: 200040] 2026-08-11-03:22:56.268.633 (EZ9999):  The error from device(chipId:0, dieId:1), serial number is 340, there is an exception of aivec error, core id is 16, error code = 0x4000000000000000, dump info: pc start: 0x124200000000, current: 0x124200000094, vec error info: 0xf01d, mte error info: 0x1d03020089, ifu error info: 0x20000ffa63dc0, ccu error info: 0x40e0080056000039, cube error info: 0, biu error info: 0, aic error mask: 0x6500020bd00028c, para base: 0x12c100000000.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:645]
       The extend info: errcode:(0x4000000000000000, 0, 0) errorStr: VEC instruction error: the ub address out of bounds. fixp_error0 info: 0x3020089, fixp_error1 info: 0x1d, fsmId:1, tslot:0, thread:0, ctxid:0, blk:0, sublk:0, subErrType:4.[FUNC:PrintCoreInfo][FILE:device_error_core_proc.cc][LINE:658]
[2026-08-11 03:22:57] ERROR: testcase failed (exit 1): gather_legacy
```

</details>

PR #1100, opened 2026-08-03
Drop a carried dependence proven absent by disjointness

Adamkh329:
This proves that case for a `pto.partition_view` whose offset is affine in the
enclosing loop's induction variable, and drops the dependence. The proof is in
address space rather than index space: disjoint index ranges imply disjoint
addresses only if the layout maps distinct index tuples to distinct addresses,
which a strided view does not promise. One iteration touches addresses within a
bounded window of its base, and the next iteration's base sits further away than
that window is wide.

The predicate refuses unless every input is a compile-time constant, and refuses
multi-buffered accesses outright: a rotating slot set has a real carried
dependence at the rotation distance, which is a different claim from no dependence
at any distance.

Measured on the sample corpus, 15 sites across 15 files. A further 15 same-pipe
barriers on 7 other kernels are refused, so the predicate discriminates rather
than firing everywhere. Effect on the rope-pack kernel, three runs each:

  A3   35.402 -> 8.980 us   (-74.6%)
  A5   68.990 -> 22.010 us  (-68.1%)

Emitted output is byte-identical on both architectures under randomised inputs
with a poisoned destination buffer.

Four refusal cases are pinned alongside the drop, because a predicate that fired
unconditionally would pass a test that only checked the drop: movement narrower
than the footprint, zero movement, movement the analysis cannot bound, and the
boundary case where movement exactly equals the footprint (adjacent, so disjoint).

### DIFF
--- a/lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp
+++ b/lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp
@@ -403,8 +403,9 @@ void InsertSyncAnalysis::InsertSync(
 //
 // An access has no loop-carried dependence on itself when it moves far enough per
 // iteration that the memory two iterations touch never meets. This proves that for
-// a `pto.partition_view` whose offset is affine in the enclosing loop's induction
-// variable, which is the shape a store walking a tensor takes.
+// a `pto.partition_view` whose offset is affine in the induction variable of the loop
+// whose back edge carries the dependence, which is the shape a store walking a tensor
+// takes.
 //
 // Every condition is cumulative and the default is to keep the sync: each helper
 // refuses the moment an input is not a compile-time constant.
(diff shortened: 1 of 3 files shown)


zhangstevenunity:
[P1] Bind the disjointness proof to the loop back edge being analyzed

`MemAnalyze` receives `forEndIndex`, but `isCarriedSelfDepAffineDisjoint` ignores it and calls `enclosingForOf(nowCompound->elementOp)`. Therefore, while analyzing an outer-loop back edge, an operation nested in another loop is still proven against the innermost `scf.for`: [relevant code](https://github.com/Adamkh329/PTOAS/blob/138ecb203a3bbf8eb774828780e5cf2342ac99ca/lib/PTO/Transforms/InsertSync/InsertSyncAnalysis.cpp#L593-L619).

A minimal trigger is:

```text
outer = 0..1
  inner = 0..7
    offset = inner * 64
    tstore 64 elements
```

Consecutive inner iterations are disjoint, but every outer iteration restarts at offset 0. The dynamic address sequence is `0, 64, ..., 448, 0, 64, ...`, so the same regions are touched again and there is a real carried WAW dependence across the outer loop.

I reproduced this with LLVM 21.1.8 using `pto-test-opt --pto-insert-sync`:

- Parent `3c4c9f7d`: one `pto.barrier <PIPE_MTE3>` is emitted before the `tstore`.
- PR head `138ecb20`: no `PIPE_MTE3` barrier is emitted inside either loop.

This removes a required ordering and can cause incorrect runtime results. Please derive/pass the exact `scf.for` represented by `forEndIndex` into the predicate (and conservatively refuse when it cannot be resolved), then add a nested-loop regression where the inner offsets are disjoint but the outer loop repeats the same ranges. The supplied single-loop FileCheck cases pass, but they do not cover this path.

PR #1101, opened 2026-08-03
Added  three absolute synchronization gates ( Correctness tests) 

Adamkh329:
Three checks over the synchronization a compilation emits, each judging one run on its own rather than against a reference:

  G3       every static event and buffer id is one the target permits for that op, direction and arch, and set/wait stay balanced per direction. Rotating ids on set_flag_dyn / wait_flag_dyn resolve at runtime and are not checked.
  
  G2       for a given id key the live intervals are pairwise disjoint. Event ids are keyed by (srcPipe, dstPipe, id) because event flags are
per-direction disjoint hardware, so EVENT_ID0 may be live in MTE2->V and V->MTE3 at once; keying on the id alone would raise a false
     violation on correct code.
     
  G1-self  every dependence the front-end analysis reports is ordered by emitted sync, derived from DepBetween directly. Pairs on mutually exclusive scf.if arms, and on A5 PIPE_V pairs the target orders itself, areexcused -- both counted and reported .

All three are read-only apart from diagnostics and off by default. They run against the existing sync path and need no new allocator.

A self-validation fixture pins that each of three fault classes is caught by exactly one gate and is silent on the other two, so no gate is redundant and none is vacuous.

Outside the new files this adds one public accessor, SyncEventIdAllocation::GetReservedEventIdNum, so the gate reads the same per-direction reservation map the sync pass obeys instead of duplicating it.

Local verification: build clean; lit 1551/1556; licence-header check passes. The 5 failures are pre-existing on unmodified main -- all under test/lit/vpto/, none referencing any flag this change adds, and this change touches no file under vpto/.

### DIFF
--- a/include/PTO/Transforms/InsertSync/SyncEventIdAllocation.h
+++ b/include/PTO/Transforms/InsertSync/SyncEventIdAllocation.h
@@ -40,6 +40,12 @@ using SyncCycle = DenseMap<int, EventCyclePool>;
  
 class SyncEventIdAllocation {
 public:
+  /// Number of event ids reserved at the tail of the intra-core pool
+  /// [0, kTotalEventIdNum) for the direction `src -> dst`; 0 when the direction
+  /// reserves nothing. Exposed so the device-id-legality gate reads the same
+  /// reservation map this allocator obeys, rather than duplicating it.
+  static uint64_t GetReservedEventIdNum(PipelineType src, PipelineType dst);
+
   SyncEventIdAllocation(SyncIRs &syncIR, SyncOperations &syncOperations)
       : syncIR_(syncIR), syncOperations_(syncOperations) {
     reserveBlockAllEventIds();
@@ -124,6 +130,7 @@ private:
   static const llvm::DenseMap<std::pair<PipelineType, PipelineType>, uint64_t>
       reservedEventIdNum;
   uint64_t reservedBlockSyncEventIdNum{0};
+
 };
  
 } // namespace pto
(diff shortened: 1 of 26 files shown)


zhangstevenunity:
Reviewed commit `2cb5faaf73741eafaa67a7385e59d0a2e5c66bd7`. I found four correctness issues that should be addressed:

1. **[P1] G1 credits conditionally executed synchronization**

   `addEdges` in `SyncOracleGates.cpp` adds coverage edges without checking control-flow dominance. In a reproducer with the barrier inside `scf.if` and the memory operations outside it, G1 reports `deps=1 covered=1`, although the false path executes both memory operations without any synchronization. Coverage should only be credited when the synchronization orders the source and destination on every relevant path.

2. **[P1] G2/G3 balance events across mutually exclusive paths**

   Event records are matched in flattened walk order. A `set_flag` in one `scf.if` branch and its `wait_flag` in the other branch pass both gates with zero violations, even though every runtime path has either a leaked set or an orphan wait. Event lifetime and balance need to be checked per CFG path.

3. **[P2] G1 does not decode legal `PipeAttr` buffer tokens**

   `computeCoverage` recognizes `PipeEventTypeAttr` and `SyncOpTypeAttr`, but `get_buf` legally accepts `PipeAttr` as part of `PTO_PipeLikeAttr`. A valid `PIPE_MTE2` to `PIPE_MTE3` buffer-token chain passes G2/G3 but is reported uncovered by G1. The PipeAttr-aware decoder already used by `bufSyncPipe` should be shared here.

4. **[P2] G1 rejects normal `PIPE_S` program order**

   The existing `test/lit/pto/scalar_gm_tstore_sync.pto` compiles normally, but enabling `--check-sync-self-coverage` fails on sequential `load_scalar` / `store_scalar` operations with an uncovered `PIPE_S -> PIPE_S` dependency. Same-pipeline scalar program ordering should be modeled according to target semantics, and this existing regression should be covered by the gate tests.

Validation: LLVM/MLIR 21.1.8; single-job native and Python package builds succeeded; all 5 newly added focused lit tests passed. The issues above were reproduced with the exact PR-built `ptoas`. `git diff --check` additionally reports a minor blank line at EOF in `SyncOracleGates.cpp`.

Issue #1148, opened 2026-08-05
[BUG] Meta issue: remaining quant performance gaps for VMI vs ASC


learning-chip:
See full report & reproducers in https://github.com/hw-native-sys/PTOAS/pull/1147

@mouliangyu

mouliangyu:
Regarding Request 2, the checked-in package is not yet sufficient to reproduce or attribute the numerical mismatch independently:

- `broken_vmi_dequant_dblbuf.ptodsl.py` and `working_vmi_dequant_narrow.ptodsl.py` are identical except for comments and the entry name. Both use the same 512-element K chunk, 256-lane halves, and `group=8`, so the advertised narrow control schedule is not actually present.
- The compile script only lowers the isolated wide/narrow `.pto` vector bodies. It does not build or run the full PTODSL kernels, and those isolated bodies do not cover grouped `vload`, e4m3-to-f32 conversion, double buffering, or writeback.
- The full PTODSL kernel also contains the Request 1 `get_block_idx`-controlled vector region, so it currently cannot complete device emission without first addressing Request 1.
- There is no device runner, deterministic input/golden data, exact `maxabs` value/first mismatching index, or complete environment and command line in the package.

Could you please add:

1. the actual narrow full PTODSL kernel that produced `maxabs = 0`;
2. a runnable wide-vs-narrow/AscendC device harness with deterministic inputs and comparison;
3. the exact shape, CANN/PTOAS commit, run commands, `maxabs`, and first mismatching element;
4. if possible, a version of the Request 2 kernel that does not depend on the Request 1 control-flow pattern.

With those pieces, we can determine whether the mismatch is in grouped load/broadcast layout, conversion, buffering, or memory movement rather than guessing from the isolated lowering result. I will handle Request 1 in a separate fix PR.

mouliangyu:
Request 1 is addressed by draft PR #1171.

The root cause was automatic VPTO vecscope inference moving `pto.get_block_idx` / `pto.get_block_num` into the inferred vector scope. That emitted `llvm.hivm.GET.BLOCK.*` inside `llvm.loop.aivector_scope`, which Bisheng could not expand. The fix keeps block queries outside the inferred scope and captures their scalar results into the guarded vector region.

Validation completed:
- all PTO lit tests: 1,684 passed, 1 unsupported;
- `current_vmi_block_idx_vf.pto`: now emits a fat object;
- `current_vmi_multibuf_expand_full.pto`: now emits a fat object.

[...]

PR #1147, opened 2026-08-05
Reproducer: remaining quant performance gaps for VMI vs ASC

learning-chip:
See `docs/repro/quant_vf_parity/FEATURE_REQUEST.md`.

Break down:
- `Request 1` is a lowering bug.
- `Request 2` is about on-device numerical error. 

Extra notes:
- This is not a PR for merge, but the reproducer for https://github.com/hw-native-sys/PTOAS/issues/1148
- This PR is on top of a small patch #1145

PR #1171, opened 2026-08-06
fix(vpto): keep block queries outside inferred vector scopes

mouliangyu:
## Summary

- keep `pto.get_block_idx` and `pto.get_block_num` outside automatically inferred VPTO vector scopes
- capture block-query results into the inferred scope so scalar bounds can still control vector bodies
- add regression coverage for block-index and block-count guards around vector load/store operations

## Root cause

Automatic vecscope inference classified pure block queries as safe scalar producers and moved them into the vector cluster when all users were also moved. This emitted `llvm.hivm.GET.BLOCK.*` inside `llvm.loop.aivector_scope`, where Bisheng failed with:

```text
Do not know how to expand the result of this operator!
```

Treating block queries as inference boundaries leaves the query outside the scope while preserving its SSA uses inside the guarded vector region.

## Validation

- `cmake --build build --target check-pto -j16`
  - 1,684 passed
  - 1 unsupported
- reproduced Request 1 before the fix with both fixtures from PR #1147
- after the fix, both fixtures complete PTOAS device emission and produce fat objects:
  - `current_vmi_block_idx_vf.pto`
  - `current_vmi_multibuf_expand_full.pto`

Addresses Request 1 of #1148.

### DIFF
--- a/lib/PTO/Transforms/PTOInferVPTOVecScope.cpp
+++ b/lib/PTO/Transforms/PTOInferVPTOVecScope.cpp
@@ -89,7 +89,11 @@ static bool isExplicitVectorScopeCarrier(Operation *op) {
 }
 
 static bool isForbiddenInsideInferredVectorScope(Operation *op) {
-  return isa<pto::VbitsortOp, pto::Vmrgsort4Op>(op);
+  // Bisheng cannot expand block-query results produced inside an AIV vector
+  // scope. Keep these scalar queries outside the inferred scope and capture
+  // their results instead.
+  return isa<pto::VbitsortOp, pto::Vmrgsort4Op, pto::GetBlockIdxOp,
+             pto::GetBlockNumOp>(op);
 }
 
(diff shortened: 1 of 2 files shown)


reedhecre:
## A5 板测失败
- 触发方式：`merged`
- 源码提交：`749fc25cf1f8`
- 日志：`/root/ptoas-board-monitor-a5/logs/20260806_203104_merged_pr1171.log`

### 失败用例
- `post_rmsnorm` (run, exit=139)

reedhecre:
## A3 板测完成（有跳过）
- 触发方式：`merged`
- 源码提交：`749fc25cf1f8`
- 结果汇总：OK `220` / FAIL `0` / SKIP `3`
- 日志：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260806_063736_merged_pr1171.log`
- 结果 TSV：`/home/zhongxuan/ptoas-board-monitor/runtime/logs/20260806_063736_merged_pr1171.tsv`

PR #1174, opened 2026-08-06
Unified allocator : Mixing bufid and event id into a unique design

Adamkh329:
This adds enable-unified-sync, an allocator that treats event ids, buffer
ids and barriers as three resource classes over a single interval-coloring
problem, instead of three separate mechanisms decided independently. It
replaces the mechanism/id-assignment stage of the existing sync pipeline
while reusing the same front-end hazard analysis, so a kernel's sync
decisions come from one coherent model rather than a chain of local choices.

 forcing the buffer-id path produces fully correct,
self-consistent emission -- hazards routed, ids assigned, get_buf/rls_buf
paired correctly, zero leaked ops -- while the same run simultaneously
assigns event ids through the ordinary path. Multiple synchronization
mechanisms coexist correctly within a single kernel, which is the core
capability this allocator is designed to unlock.

On the current sample corpus, coverage and op-count parity with the
existing sync path hold exactly, and the absolute coverage gate reports
identical results under both paths -- this PR changes how synchronization
is decided, not what gets emitted for kernels the existing corpus already
represents.

Where this goes next: as kernels that actually pressure the id pools enter
the corpus (multi-buffered or high-occupancy workloads), the buffer-id
routing path built here is what starts engaging automatically -- no
allocator changes needed, just corpus growth. Follow-up work will extend
the reachability analysis this allocator shares with the sync gates to
reason about loop trip counts, and settle an open question about same-pipe
ordering guarantees on A3 that affects how conservative the gates need to
be.

Note: next focus is on the synchronization points themselves -- converting
event-id-based sync to buffer-id sync specifically across if/else scope
boundaries, where a hazard's set and wait can sit on opposite branches.

Update: fixes all six correctness issues raised in review, plus three earlier commits kept in and named below.

Base: f8515c33. 11 commits, 20 files, +873/-47.

What changed:

1. 150ac571 - a dependency that spanned both a converted and unconverted buffer was being converted anyway, silently dropping the ordering on the unconverted side. That case is now rejected outright instead of being partially applied.

2. de48e6d8 - this one turned out to already be fixed by earlier work on the branch. Added a regression test so it stays fixed.

3. 9e06a9d2 - several failure conditions (running out of id capacity, invalid nesting, generation failure) were being computed but only printed in debug mode, so a real failure could pass silently in normal builds. These now stop the build with a real error.

4. 2c12592c - one synchronization mode was missing a check the others already had, so it accepted an operation it shouldn't have. Now rejected consistently across all modes, with a regression test.

5. f4ce8ffd - the mechanism comparing two compiler runs for consistency used a hash value that isn't guaranteed to be the same between runs, so identical output could be flagged as different. Replaced with a stable hash, and added a version tag so an old saved result is recognized as outdated rather than compared incorrectly.

6. a41229ee - a debug check was firing and crashing the program before the intended error message could be printed. Removed the check; the real error now prints as intended.

Also included, from before this review started:
- 63073275 - fixes a case where buffer synchronization was being computed from stale information, which could place it incorrectly.
- d33e5843 - records when two allocations end up sharing the same memory address, for visibility only; doesn't change behavior.
- 5e31dbea - adds an optional report listing those shared-address cases, off by default.

Verification, on two build configurations (normal, and with debug checks enabled):

                        normal build          debug-checks build
tests                   1664 passed, 0 failed 1664 passed, 0 failed
ctest                   50/50                 50/50
filtered suite          60 passed, 0 failed   60 passed, 0 failed

Worth knowing:
- #1's fix has no current input that actually triggers it - checked carefully, this is because the situation it guards against can't currently occur, not because it wasn't tested properly.

- The new test for #2 pins real behavior but isn't proven to catch the exact regression it's meant to guard against.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,10 @@
 build/
 build_plain/
 build_plan/
+# Any out-of-tree build directory, however named. `build/` alone let a second
+# configuration (build-assert/) be added by `git add -A`.
+build*/
+build-assert/
 install/
 
 # TileLang ST standalone build outputs (see temp_docs/standalone_st.md)
--- a/docs/designs/ptoas-unified-sync-design.md
+++ b/docs/designs/ptoas-unified-sync-design.md
@@ -1,0 +1,493 @@
+# PTOAS Unified Intra-Core Sync Allocator Design
+
+## Intent
+
+`--enable-unified-sync` allocates intra-core synchronization for one function
+across all three hardware mechanisms from a single model, instead of committing to
+a mechanism before the cost of that choice is known.
+
+The existing passes each own one mechanism. `--enable-insert-sync` allocates event
+ids and falls back to a `PIPE_ALL` barrier when a pool is exhausted.
+`--enable-bufid_sync` allocates buffer-id tokens (see `docs/bufid_sync_a5_design.md`).
+Neither can trade one resource against the other: when a pipe direction runs out of
+event ids, insert-sync must serialize with a barrier even where buffer-id tokens are
+free.
+
+This pass makes the mechanism a decision rather than a premise. Hazards whose
+direction would overflow its event pool are routed to buffer-id tokens where the
+buffer is eligible; what remains is coloured into the event pools by live interval;
+anything that still does not fit spills to a barrier.
+
+The dependence analysis is **shared** with `pto-insert-sync` rather than
+reimplemented. Only the assignment of mechanisms and ids differs. That is a
+deliberate constraint, not a convenience: it means both passes see the same program
+and the same hazards, which is what makes a differential comparison between them
+meaningful rather than a comparison of two front ends.
+
+Non-goals:
+
+- Cross-core synchronization. `pto.sync.set` / `pto.sync.wait` carry FFTS or
+  intra-block flag ids from a different id space, and are untouched.
+- Replacing any existing pass. This is one more automatic-synchronization mode,
+  mutually exclusive with the other four and off by default. Passing more than one is
+  rejected with a diagnostic rather than resolved by precedence.
+- Scheduling. The pass does not move compute ops; it only decides how the orderings
+  the front end reports are enforced.
+
+## Mechanisms
+
+| mechanism | ops | pool | cost |
+|---|---|---|---|
+| event id | `pto.set_flag` / `pto.wait_flag` | 8 per `(src, dst)` pipe direction, disjoint hardware per direction | directed: orders one direction |
+| buffer id | `pto.get_buf` / `pto.rls_buf` | one pool of K, shared across directions; K=0 on A3, K=32 on A5 | a ticket protocol on one buffer; co-consumers of one token serialize |
+| barrier | `pto.barrier <PIPE_ALL>` | unlimited | orders every anchor pair spanning it |
+
+Event pools are per direction and physically disjoint: id 0 in `MTE2->V` is a
+different flag from id 0 in `V->MTE3`. The buffer-id pool is not -- it is one global
+K, so a buffer-id token consumed anywhere reduces what is available everywhere.
+
+A hazard always ends on some mechanism. There is no resting state in which a
+hazard has been decided but carries no resource; `SyncCodegen` fails the pass if a
+set/wait pair reaches emission with no event id.
+
+## Enabling the pass
+
+```bash
+ptoas input.pto --pto-arch=a5 --enable-unified-sync -o output.cpp
+```
+
+`--enable-unified-sync` is mutually exclusive with the other four automatic modes --
+`--enable-insert-sync`, `--enable-bufid_sync`, `--enable-inject-barrier-all-sync` and
+`--enable-graph-sync-solver` -- and passing two is an error rather than a precedence
+question. There is no compiler default: a run that passes none of them inserts no
+synchronization at all, which is why every measurement in this document names its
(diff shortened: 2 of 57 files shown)


zhangstevenunity:
I reviewed exact head `f8515c334e2f9e75cac5ec4fc50982450026816a` locally with the LLVM/MLIR 21.1.8 assertion build. I found six blocking correctness issues:

1. **[P1] Reject split hazards before routing them to buffer IDs.**

   [`routeBuffers`](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/UnifiedSyncModel.cpp#L528-L547) detects a hazard spanning routed and unrouted buffer clusters, but the caller ignores `splitHazards` and [routes the entire hazard when any cluster matches](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/PTOUnifiedSync.cpp#L291-L307). This suppresses event synchronization for the unrouted portion without providing its buffer-token synchronization. Routing should be closed over each buffer-hazard component, or any plan with `splitHazards != 0` should fail before mutating hazards.

2. **[P1] Preserve routed logical IDs across same-pipe merging.**

   Routed IDs are [captured before buffer analysis](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/PTOUnifiedSync.cpp#L291-L295), but [`optimizeSamePipeMerge()` rewrites logical IDs](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/BufidSync/BufidSyncAnalysis.cpp#L567-L609) before the old routed-ID set is used to [filter token operations](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/PTOUnifiedSync.cpp#L505-L528). Merging a routed and an unrouted ID can either remove required `get_buf`/`rls_buf` operations or apply tokens to a dependency whose event synchronization remains. The merge map needs to be propagated into the routed set, or filtering must happen before IDs are rewritten.

3. **[P1] Propagate terminal buffer-allocation and codegen failures.**

   The unified path computes `needsReuse()`, `nestOk`, and `emitOk` but [continues regardless of their final values](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/PTOUnifiedSync.cpp#L530-L547). The production buffer-ID pass explicitly [fails for exhausted IDs, invalid nesting, and codegen failure](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/BufidSync/BufidSyncPass.cpp#L95-L121). The unified path can therefore report success after producing out-of-range IDs, invalid token nesting, or partial synchronization code.

4. **[P1] Reject `pto.tassign` with unified sync.**

   The CLI [rejects `tassign` with the other automatic synchronization modes](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/tools/ptoas/ptoas.cpp#L3366-L3404), but omits `enableUnifiedSync`. On `tassign_level3_loop_rebind.pto`, `--enable-insert-sync` exits with the expected incompatibility diagnostic, while `--enable-unified-sync` silently exits 0. Unified sync uses the same memory-analysis path, which does not model dynamic address rebinding, so this needs the same guard and a regression test.

5. **[P1] Use a stable digest for persisted coverage profiles.**

   [`computeCoverage` stores `llvm::hash_value(signatureText)`](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/SyncOracleGates.cpp#L1180-L1182), then compares the value in another `ptoas` process. LLVM explicitly defines this hash as unstable across processes/executions. Consequently, identical source is rejected as an `anchor-signature-mismatch`. This directly fails `sync_gate_g1_coverage.pto` and `sync_forced_overflow.pto`. Please use a stable digest such as xxHash, or persist the normalized signature itself.

6. **[P1] Do not assert before emitting the expected codegen diagnostic.**

   [`SyncCodegen::SyncInsert`](https://github.com/hw-native-sys/PTOAS/blob/f8515c334e2f9e75cac5ec4fc50982450026816a/lib/PTO/Transforms/InsertSync/SyncCodegen.cpp#L299-L307) executes `assert(false)` before emitting `error: sync codegen`. In the repository's assertion build, the PR's own negative test aborts instead of producing the expected diagnostic. The implementation should emit the diagnostic and propagate pass failure without an unconditional assertion.

Validation performed:

- LLVM/MLIR 21.1.8 assertion build, `PTOASPythonPackage`: 320 build steps passed with `-Werror`.
- `llvm-lit -v .../test/lit --filter sync_`: 54 passed, 3 failed out of 57.
- Failed tests: `sync_mechanism_field.pto`, `sync_gate_g1_coverage.pto`, and `sync_forced_overflow.pto`.
- `git diff --check upstream/main...HEAD`: passed.
- GitHub currently reports no remote checks for this PR.

PR #1269, opened 2026-08-17
fix(ptodsl): preserve entering SSA values for nested partial assignments

### LINKED ISSUE #1252: [Bug] [PTODSL] [阻塞] 嵌套 Python if 的部分赋值错误覆盖未赋值的 ui32 loop carry，导致数值语义错误
### Component

Python bindings (python/)

### Description

PTODSL AST rewrite 在处理循环内嵌套 Python if 时，没有正确保留分支中未赋值变量的 incoming value。
当循环携带 hi_bits、lo_bits 两个 ui32，最内层 else 只更新：hi_bits = q1_bits
按 Python 语义，lo_bits 应保持进入该分支前的值。但生成的 scf.if 将 lo_bits 错误替换成了相邻分支的 q2_bits。
生成的 MLIR 可以通过 verifier 和后端编译，因此该问题不会表现为编译错误，而是在运行时产生静默数值错误。
该问题可能与已关闭的 #1073/#1095 属于同类控制流 SSA 问题，但在“循环 carry + 多层嵌套 if + partial assignment”场景下仍可复现。

### Reproduction (minimal)

```markdown
from ptodsl import pto


@pto.jit(target="a5", backend="vpto", mode="explicit")
def nested_partial_assign_probe(
    rows: pto.i32,
    cond3: pto.i1,
    cond2: pto.i1,
    cond1: pto.i1,
    out: pto.ptr(pto.ui32, "gm"),
):
    lo_bits = pto.const(0, dtype=pto.ui32)
    hi_bits = pto.const(0x40000000, dtype=pto.ui32)

    q1_bits = pto.const(0x10000000, dtype=pto.ui32)
    q2_bits = pto.const(0x20000000, dtype=pto.ui32)
    q3_bits = pto.const(0x30000000, dtype=pto.ui32)

    for _ in range(rows):
        if cond3:
            lo_bits = q3_bits
        else:
            if cond2:
                lo_bits = q2_bits
                hi_bits = q3_bits
            else:
                if cond1:
                    lo_bits = q1_bits
                    hi_bits = q2_bits
                else:
                    hi_bits = q1_bits  # lo_bits must remain unchanged

    pto.stg(hi_bits, out, 0)
    pto.stg(lo_bits, out, 1)


print(nested_partial_assign_probe.compile().mlir_text())
```

### Expected behavior

scf.yield %q1_bits, %arg15 : ui32, ui32

### Actual behavior / error logs

```text
实际 TileLang top-p kernel 中，循环参数为scf.for ... iter_args(%arg14 = %hi_bits, %arg15 = %lo_bits)
最内层 else 只更新 hi_bits = q1_bits，但生成：scf.yield %q1_bits, %q2_bits : ui32, ui32
第二个结果错误地使用了 q2_bits。
```

### Git commit

0eb64461e0c1321183a927fbe744dcfb9a01c7aa

### Host platform

None

### Target Ascend arch (if relevant)

a5

### PTOAS build level (if relevant)

None

HecreReed:
Fixes #1252

## Problem

PTODSL rewrites nested Python conditionals into scf.if structures while
tracing. Both dynamic sibling branches share the same Python variable
environment: the branch traced first rebinds a Python variable, and the
later branch observes that rebinding. For a variable a branch does not
assign (partial assignment), the rewritten fallback captured the current
Python value instead of the entering value, silently substituting the
sibling branch value in the merged yields:

    scf.yield %q1_bits, %arg15 : ui32, ui32   (expected)
    scf.yield %q1_bits, %q2_bits : ui32, ui32 (actual, wrong lo_bits)

The generated MLIR passes the verifier and the backend, so it surfaced as
a silent numerical error at runtime (reported from a real TileLang top-p
kernel). A second gap blocked the minimal reproducer: a plain runtime for
loop only recognized explicitly-read loop carries, so the variable kept
only via the partial-assignment default path was rejected as
last-iteration-only before the wrong yield could ever be produced.

## Changes (PTODSL AST rewrite only)

- _ControlFlowRewriter._rewrite_if: compute the names that need their
  entering value (fallbacks plus names live-in to a branch), snapshot those
  values once on the runtime path right before pto.if_, and restore the
  Python binding at the start of both dynamic branch bodies. The restore
  assignments emit no IR; they only isolate sibling branch tracing
  environments. _branch_assign now reuses the entry snapshot for unassigned
  names. The native Python bool path is unchanged.
- _read_before_assignment_names: accept an optional live_after seed;
  _rewrite_for seeds it with the body stores that are live after the loop so
  partial-assignment default paths are recognized as loop carries.
  Accumulator loops keep working and last-iteration-only diagnostics are
  preserved.

## Tests

Two focused probes in ptodsl/tests/test_jit_compile.py with MLIR op-tree
SSA identity assertions:

- no-loop nested partial assignment: the inner else must yield the pre-if
  value, not the sibling branch rebinding;
- the #1252 runtime loop structure: three nested scf.if inside an scf.for
  carrying hi_bits/lo_bits, asserting the innermost else yields the lo_bits
  loop-carried block argument (and not the q2 constant).

Both modules pass parse round-trip and the verifier.

## Verification (LLVM 19 toolchain)

    cmake --build build --target PTODSLPackage PTOAS_Python
    ctest --test-dir build --output-on-failure -R '^(ptodsl_jit_compile|ptodsl_section|ptodsl_jit_diagnostics)$'
    cmake --build build --target check-dsl

All 39 PTODSL ctest targets pass, including the new probes (RED verified on
the unfixed baseline: the innermost-else yield carried the sibling branch
value).

### DIFF
--- a/ptodsl/ptodsl/_ast_rewrite.py
+++ b/ptodsl/ptodsl/_ast_rewrite.py
@@ -2484,7 +2484,7 @@ class _ControlFlowRewriter:
         if unsupported_last_values:
             raise PTODSLAstRewriteError(
                 "ast_rewrite=True runtime while cannot expose last-iteration-only values yet; "
-                f"use explicit pto.while_(...).carry(...) for {unsupported_last_values}"
+                f"use explicit pto.while_(...) state for {unsupported_last_values}"
             )
         if not carry_names and not (control["break"] or control["continue"] or stmt.orelse):
             raise PTODSLAstRewriteError(
--- a/ptodsl/ptodsl/_ast_rewrite.py
+++ b/ptodsl/ptodsl/_ast_rewrite.py
@@ -1759,6 +1759,16 @@ class _ControlFlowRewriter:
                     next_static_iters[stmt.target.id] = values
             saved_control_stack = self._loop_control_stack
             self._loop_control_stack = []
+            name_target = stmt.target.id if isinstance(stmt.target, ast.Name) else None
+            entry_with_iv = set(bound_before or ())
+            if name_target is not None:
+                entry_with_iv.add(name_target)
+            orelse_with_iv = set(bound_before or ())
+            # The induction target is only definitely bound in the else clause
+            # when the static range is known non-empty; an empty static range
+            # jumps straight to the else clause with the target unbound.
+            if name_target is not None and values is not None and len(values) > 0:
+                orelse_with_iv.add(name_target)
             try:
                 stmt.body = self.rewrite_block(
                     stmt.body,
@@ -1766,7 +1776,7 @@ class _ControlFlowRewriter:
                     live_after_slots=live_after_slots,
                     allow_loop_control=True,
                     static_iters=next_static_iters,
-                    bound_on_entry=set(bound_before or ()) | {stmt.target.id},
+                    bound_on_entry=entry_with_iv,
                 )
                 stmt.orelse = self.rewrite_block(
                     stmt.orelse,
@@ -1774,7 +1784,7 @@ class _ControlFlowRewriter:
                     live_after_slots=live_after_slots,
                     allow_loop_control=True,
                     static_iters=static_iters,
-                    bound_on_entry=set(bound_before or ()) | {stmt.target.id},
+                    bound_on_entry=orelse_with_iv,
                 )
             finally:
                 self._loop_control_stack = saved_control_stack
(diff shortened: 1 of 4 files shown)


ptodsl/ptodsl/_ast_rewrite.py
```diff
@@ -2053,7 +2254,7 @@ def visit_Name(inner, node):
         return result
 
     def _rewrite_while(self, stmt, *, live_after, live_after_slots=None,
-                       allow_loop_control=False, static_iters=None):
+                       allow_loop_control=False, static_iters=None, bound_before=None):
```
Zhendong404, line 2496: The for-loop paths (`_rewrite_for`, `_rewrite_controlled_for`) now gate implicit loop carries on a definite binding before the loop via `bound_before`, but `_rewrite_while` keeps the old criterion a few lines below: `carry_names = body_info.stores & (test_info.loads | body_info.loads | set(live_after))`, with neither the `live_after` liveness seeding nor the definite-binding gate.

Two consequences:

1. `for` and `while` now apply different standards to the same "previous value kept via a partial-assignment default path" semantics this PR is fixing.
2. A name that is live after the `while` loop but not bound before it is still carried, so entering the loop reads an unbound Python local at `pto._while(...)` setup — surfacing as a raw `NameError` instead of the clear last-iteration-only diagnostic the for path now produces.

Suggest aligning the while path (seed backward liveness with `assigned_live_after` and gate implicit carries on `bound_before`), or at least aligning the diagnostic. Fine as a follow-up if you prefer to keep this PR scoped.

Zhendong404 (approved):

PR #1314, opened 2026-08-21
add sync.set <PIPE_MTE1> / sync.set <PIPE_V>

Likai-19:
Summary
Add PTODSL sync facades that lower directly to the unified pto.sync.set / pto.sync.wait ops, and widen the A5 static event-id range for those ops from [0, 15] to [0, 31].

Problem
A5 mixed kernels need intra-block cross-pipe synchronization between the Cube side and the Vector side. In many schedules the natural signal is emitted directly by a producer pipe such as PIPE_MTE1 or PIPE_V, and the consumer waits on that same signal. The previous state made this difficult for two reasons:

The PTODSL surface exposed only set_intra_block / wait_intra_block style facades, which lower to the named block-sync ops. There was no public PTODSL facade for the unified pto.sync.set / pto.sync.wait ops, even though the lower-level VPTO dialect already supported them.
The A5 verifier restricted static sync.set / sync.wait event ids to [0, 15], while physical intra-block event ids in [16, 31] are valid.
As a result, kernels that wanted to signal directly from a non-FIX/non-MTE3 pipe were forced to insert bridge signals through an intermediate pipe. Those bridges introduce extra set_flag / wait_flag pairs and extra sync.wait events, which can amplify the wait count on a single pipe and serialize work that should otherwise overlap. This is a common issue for mixed kernels, not specific to any particular workload.

Fixes: https://github.com/mouliangyu/PTOAS/issues/576

### DIFF
--- a/ptodsl/ptodsl/_ops.py
+++ b/ptodsl/ptodsl/_ops.py
@@ -6811,6 +6811,54 @@ def wait_intra_block(pipe, event_id):
     _pto.wait_intra_block(_pipe_attr(pipe), event_operand)
 
 
+def set_cross_flag(pipe, event_id):
+    """``pto.set_cross_flag(pipe, event_id)`` – cross-core sync facade for ``pto.sync.set``."""
+    _validate_sync_pipe(pipe, context="set_cross_flag(pipe, event_id)", allowed=("PIPE_FIX",))
+    event_operand = _sync_event_id_operand(event_id, context="set_cross_flag(..., event_id=...)")
+    _pto.sync_set(_pipe_attr(pipe), event_operand)
+
+
+def wait_cross_flag(pipe, event_id):
+    """``pto.wait_cross_flag(pipe, event_id)`` – cross-core sync facade for ``pto.sync.wait``."""
+    _validate_sync_pipe(pipe, context="wait_cross_flag(pipe, event_id)", allowed=("PIPE_FIX",))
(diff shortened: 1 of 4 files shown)


ptodsl/ptodsl/_ops.py
```diff
@@ -6811,6 +6811,54 @@ def wait_intra_block(pipe, event_id):
...
+    event_operand = _sync_event_id_operand(event_id, context="wait_cross_flag(..., event_id=...)")
+    _pto.sync_wait(_pipe_attr(pipe), event_operand)
+
+
+def set_intra_flag(pipe, event_id):
+    """``pto.set_intra_flag(pipe, event_id)`` – intra-block sync facade for ``pto.sync.set``."""
+    _validate_sync_pipe(
+        pipe,
+        context="set_intra_flag(pipe, event_id)",
+        allowed=("PIPE_FIX", "PIPE_MTE1", "PIPE_MTE2", "PIPE_MTE3", "PIPE_V"),
```
mouliangyu, line 6833: [P2] A5 头文件确实使用 `PIPE_MTE1`/`PIPE_V` 以及物理 event id 16..31，因此 verifier 放宽范围的方向是合理的。但当前测试没有实际覆盖这些新路径：没有 MTE1/V producer-consumer 的 runtime/Camodel case，也没有 event id 16 或 31 的生成和执行验证。建议至少补一个 mixed Cube/Vector sync case，验证生成的 `sync.set`/`sync.wait` 在 simulator 或目标后端上的实际配对语义。

PR #1330, opened 2026-08-24
Enable VPTO soft post-update by default

yexiaosu:
Enable the VPTO soft post-update optimization by default.

The optimization can still be disabled with:

`--enable-vpto-soft-postupdate=false`

Updated related documentation and regression tests.


mouliangyu (approved):
已完成代码检视与本地回归，未发现阻塞问题。相关 soft_postupdate lit 测试及 49 个回归用例通过，批准合入。

PR #1362, opened 2026-08-27
fix: address PTOAS codecheck findings

jimmychou0:
## Summary

- refactor oversized and high-complexity Python and C++ routines reported by CodeCheck
- extract shared PTOAS compile, EmitC, ELF-patching, and PTOBC codec helpers; table-drive the ptobc opcode lookups (v0 wire payloads unchanged)
- fix validation and control-flow issues found while resolving the findings
- resolve the previously deferred structural findings with equivalence-preserving splits:
  - `ptodsl/ptodsl/_ops.py` (5913 nbnc) → eight domain modules (`_ops_common/_core/_vmem/_mask/_vmath/_tile/_mte/_simt`) plus a re-export shim; `__all__` and every module-level name (public and private, incl. the `lib/TileOps` private-name imports) stay importable from `ptodsl._ops`; `test_vector_cube_ops.py` patch sites now use a `patch_ops` helper that preserves the pre-split mock seam
  - `tools/ptoas/ptoas.cpp` (3527 nbnc) → four translation units (`ptoas.cpp` ~333, `ptoas_name_hints.cpp` ~826, `ptoas_cpp_rewrite.cpp` ~1208, `ptoas_pipeline.cpp` ~1399 nbnc) coordinated by `ptoas_internal.h` (shared constants/aliases, extern cl::opt declarations, cross-TU entry points)
  - `compilePTOASModule` cyclomatic complexity: the branch-dense clusters inside `runMainLoweringPipeline` extract into `appendFusionFrontendPasses` / `appendPlanMemoryPasses` / `appendAutoSyncPasses` with no pass-ordering or gating changes
- preserve public ABI, CLI entry points, cross-TU APIs, and generated PTOBC opcode payloads

## Validation (144, LLVM 19.1.7)

- Full rebuild: `PTOASCompiler`, `PTOASPythonCore`, `PTODSLPackage`, `PTOASPythonPackage`, `ptobc`
- CTest: 68/68 passed (45 PTO/PTODSL, 23 PTOBC)
- lit `check-pto`: 1843 passed / 1 platform-unsupported
- Smoke: `--emit-pto-ir`, EmitC C++, VPTO emission, VPTO TADDS for A5
- Import smoke incl. `TileOps.a5.thistogram` private-name imports
- changed-code compliance check: 0 genuine errors (remaining G.FMT.11 prefilter hits are false positives on already-braced nested-paren conditions; the one assert is a debug-only internal invariant)
- `python3 -m compileall -q ptodsl/ptodsl python/pto tools`; `git diff --check`


tools/ptobc/generated/ptobc_opcodes_v0.h
```diff
@@ -258,438 +257,75 @@ inline const OpInfo *lookupByOpcode(uint16_t opcode) {
...
-    .Case("pto.tmatmul.bias",
-          OpcodeAndVariant{0x1032, kHasVariant, kVariantBias})
-    .Case("pto.tmatmul.mx",
-          OpcodeAndVariant{0x1033, kHasVariant, kVariantDefault})
-    .Case("pto.tmatmul.mx.acc",
-          OpcodeAndVariant{0x1033, kHasVariant, kVariantAcc})
-    .Case("pto.tmatmul.mx.bias",
-          OpcodeAndVariant{0x1033, kHasVariant, kVariantBias})
-    .Default(std::nullopt);
+  for (const OpInfo &entry : kOpTable) {
```
Zhendong404, line 308: Behavior change vs. the old `StringSwitch`: the old `lookupOpcodeAndVariantByFullName` deliberately did *not* list `pto.section` (see the NOTE two lines above — it is not a real op name; only `pto.section.cube`/`.vector` are registered), so a bare `"pto.section"` fell through to `std::nullopt`. The new table-driven loop matches the `kOpTable` entry `{0x0006, "pto.section", has_variant=1, ...}` and returns `{0x0006, 1, 0}`, i.e. it silently resolves to the cube variant.

This is unreachable for registered ops (the encoder dispatches on `op.getName()`), but `runEncode` sets `allowUnregisteredDialects(true)`, so an input containing an unregistered op literally named `pto.section` would previously fall through to the generic-record path (or hard error), and is now compact-encoded as known-op `0x0006` cube. If strict no-behavior-change is intended, suggest excluding `pto.section` from the name-based match (e.g. an explicit `if (fullName == "pto.section") return std::nullopt;`, or skipping `has_variant` base names in this loop). Same root cause also makes `pto.tscatter.maskpattern` resolvable by name where it previously was not (lower impact — that lookup path has no registered-op caller).

jimmychou0: Good catch — fixed in b13300e83. Both `pto.section` and `pto.tscatter.maskpattern` now return `std::nullopt` via an explicit guard before the kOpTable scan, restoring the original switch's fall-through to the generic record.

I also re-ran a bidirectional equivalence check against the original per-name switch (all 184 names, both directions plus value equality): the lookup is now exactly equivalent — no newly-resolvable names, no dropped names, no mapping diffs. The base-name → base-opcode mappings (pto.textract/tinsert/tmov/tstore/tfillpad → 0x1021/0x102D/0x1038/0x1065/0x1023) are unchanged from the original switch, so those were not affected.

Validated on 144: ptobc rebuilt, ctest -R ptobc 23/23 passing.

PR #1408, opened 2026-08-31
fix(ptodsl): preserve signless arithmetic operand types

liuzidi:
Fix runtime integer arithmetic lowering after 5eb87c21. Integer literals are emitted signless, and mixed signed/signless operands are reconciled in signless arithmetic types. Explicit signed/unsigned scalar and VMI storage semantics remain unchanged. Regression: PTODSL jit compile suite and topk_gate E=256/384/512/72 K=8/9/8/6 num_sms=2.

### DIFF
--- a/ptodsl/ptodsl/_scalar_adaptation.py
+++ b/ptodsl/ptodsl/_scalar_adaptation.py
@@ -15,6 +15,7 @@ from ._types import (
     _restore_integer_signedness,
     _signless_integer_type,
     _strip_integer_signedness,
+    _parse_integer_value,
 )
 
 from ptoas.mlir.dialects import arith
@@ -180,15 +181,15 @@ def reconcile_typed_runtime_binary_operands(lhs, rhs):
     rhs_type = rhs.type
 
     if lhs_type == rhs_type:
-        return lhs, rhs, classify_runtime_scalar_type(lhs_type)
+        return lhs, rhs, classify_runtime_scalar_type(lhs_type), lhs_type
 
     if IndexType.isinstance(lhs_type) and IntegerType.isinstance(rhs_type):
         rhs = arith.IndexCastOp(IndexType.get(), _strip_integer_signedness(rhs)).result
-        return lhs, rhs, "index"
+        return lhs, rhs, "index", lhs_type
 
     if IntegerType.isinstance(lhs_type) and IndexType.isinstance(rhs_type):
         lhs = arith.IndexCastOp(IndexType.get(), _strip_integer_signedness(lhs)).result
-        return lhs, rhs, "index"
+        return lhs, rhs, "index", rhs_type
 
     if IntegerType.isinstance(lhs_type) and IntegerType.isinstance(rhs_type):
         lhs_width = IntegerType(lhs_type).width
@@ -200,11 +201,14 @@ def reconcile_typed_runtime_binary_operands(lhs, rhs):
         # restore the authored signedness only on the result.  This avoids
         # manufacturing a signed literal cast that can be reused with stale
         # type metadata by the tracer.
+        authored_type = lhs_type
+        if _integer_signedness(lhs_type) == "signless" and _integer_signedness(rhs_type) != "signless":
+            authored_type = rhs_type
         if _integer_signedness(lhs_type) == "signless" or _integer_signedness(rhs_type) == "signless":
             target_type = _signless_integer_type(target_type)
         lhs = coerce_integer_like(lhs, target_type)
         rhs = coerce_integer_like(rhs, target_type)
-        return lhs, rhs, "integer"
+        return lhs, rhs, "integer", authored_type
 
     raise TypeError(
         "runtime scalar operators require matching scalar types or an index/integer pair; "
@@ -266,7 +270,7 @@ def _materialize_runtime_literal(value, anchor_type):
     # stripping it again (the latter used to print malformed ``siN to iN``
     # casts for expressions such as ``num_tokens + 3``).
     signless_type = _signless_integer_type(anchor_type)
-    return arith.ConstantOp(signless_type, int(value)).result
+    return arith.ConstantOp(signless_type, _parse_integer_value(value, target_type=anchor_type)).result
 
 
 def _coerce_float_like(value, target_type):
--- a/ptodsl/ptodsl/_scalar_adaptation.py
+++ b/ptodsl/ptodsl/_scalar_adaptation.py
@@ -194,6 +194,14 @@ def reconcile_typed_runtime_binary_operands(lhs, rhs):
         lhs_width = IntegerType(lhs_type).width
         rhs_width = IntegerType(rhs_type).width
         target_type = lhs_type if lhs_width >= rhs_width else rhs_type
+        # Arithmetic operands are signless.  When a signed/unsigned runtime
+        # value is combined with a literal (which is intentionally emitted as
+        # signless), keep the common operand type signless through the op and
+        # restore the authored signedness only on the result.  This avoids
+        # manufacturing a signed literal cast that can be reused with stale
+        # type metadata by the tracer.
+        if _integer_signedness(lhs_type) == "signless" or _integer_signedness(rhs_type) == "signless":
+            target_type = _signless_integer_type(target_type)
         lhs = coerce_integer_like(lhs, target_type)
         rhs = coerce_integer_like(rhs, target_type)
         return lhs, rhs, "integer"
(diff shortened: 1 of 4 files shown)


liuzidi:
Validation update: with PTOAS base 5eb87c21, full ptodsl/tests/test_jit_compile.py passes. TileLang topk_gate representative failing configurations (E/K 256/8, 384/9, 512/8, 72/6; num_sms=2) all complete PTO lowering. cast_back packed UE8M0 with tile_k=256 (8-byte logical row stride) and tile_k=512 (16-byte logical row stride), hidden=2048/4096, also compile without kernel padding, confirming physical-access legalization.

liuzidi:
Regression test now asserts no malformed literal cast (`%c*_i32 = unrealized_conversion_cast ... si32 to i32`) and passes after rebuilding PTOAS Python sources.

liuzidi:
Full current-state audit: all regenerated topk_gate cases in Failures (num_groups=1, E=256/384/512/72, T=4/512/4001, K=8/9/6, num_sms=2) pass with this branch. Three old /tmp PTO artifacts from before the fix still fail if replayed; they are stale and were not regenerated. Fresh artifacts pass. No kernel padding changes are present.

liuzidi:
Additional boundary validation: explicit si32/ui32 mixed arithmetic probe passes and preserves arith.maxsi/minui plus signed GM store. This confirms the signless-literal branch does not alter explicit signed/unsigned operand semantics.

liuzidi:
Final audit rerun on clean rebuilt sources: full test_jit_compile.py PASS; dedicated issue_1405 regression PASS; topk_gate E/K 256/8, 384/9, 512/8, 72/6 with num_sms=2 PASS; cast_back original kernel tile_k=256/512 (8/16-byte logical SF row stride), hidden=2048/4096 PASS. No kernel padding changes.

liuzidi:
Revalidated against unchanged origin/main=5eb87c21a. Rebuilt PTOAS and reran test_jit_compile.py, test_scalar_cast.py, test_ptoas_frontend_verify.py: all PASS. PR remains minimal (only _scalar_adaptation.py + dedicated regression test), no kernel edits.

liuzidi:
Additional width regression: generated and verified signed/unsigned runtime arithmetic probes for si8/ui8, si16/ui16, si32/ui32, si64/ui64. All compile with arith.addi and typed stores, confirming the fix is not si32-specific.

ptodsl/ptodsl/_scalar_adaptation.py
```diff
@@ -194,6 +194,14 @@ def reconcile_typed_runtime_binary_operands(lhs, rhs):
...
         rhs_width = IntegerType(rhs_type).width
         target_type = lhs_type if lhs_width >= rhs_width else rhs_type
+        # Arithmetic operands are signless.  When a signed/unsigned runtime
+        # value is combined with a literal (which is intentionally emitted as
+        # signless), keep the common operand type signless through the op and
+        # restore the authored signedness only on the result.  This avoids
+        # manufacturing a signed literal cast that can be reused with stale
+        # type metadata by the tracer.
+        if _integer_signedness(lhs_type) == "signless" or _integer_signedness(rhs_type) == "signless":
+            target_type = _signless_integer_type(target_type)
```
Zhendong404, line 208: This rule silently changes unsigned semantics. Since `_materialize_runtime_literal` now anchors literals as signless, any `uiN` runtime value combined with a literal lands here and the result type becomes signless iN. Downstream op selection reads signedness from the reconciled type (`_runtime_scalar_ops.py`: `emit_runtime_binary_op`/`emit_runtime_compare`/`emit_runtime_min`/`emit_runtime_max`), where signless defaults to signed — so e.g. `(n_ui32 + 3) // 4` now lowers to `floordivsi` and `n_ui32 + 3 < k` to `slt` instead of `divui`/`ult`. For values >= 2^31 this is a silent numeric wrong-result, not a compile error. Suggest preserving the non-signless side's authored signedness on the result (compute in the signless domain, restore the uiN/siN identity afterwards), or at minimum capture the operand signedness before reconciliation for op selection. Either way, please add a ui32 regression test.

Zhendong404 (commented):
Overall the direction is right: keeping arith operands signless from materialization through the op removes the "si32 identity + i32 value" mismatch that broke topk_gate. A few issues inline — the main one is a silent unsigned semantics regression introduced by the signless-reconcile rule.

ptodsl/ptodsl/_scalar_adaptation.py
```diff
@@ -194,6 +194,14 @@ def reconcile_typed_runtime_binary_operands(lhs, rhs):
         lhs_width = IntegerType(lhs_type).width
         rhs_width = IntegerType(rhs_type).width
         target_type = lhs_type if lhs_width >= rhs_width else rhs_type
+        # Arithmetic operands are signless.  When a signed/unsigned runtime
+        # value is combined with a literal (which is intentionally emitted as
+        # signless), keep the common operand type signless through the op and
+        # restore the authored signedness only on the result.  This avoids
```
Zhendong404, line 201: This comment says "restore the authored signedness only on the result", but the implementation does not actually restore the authored signedness: `emit_runtime_binary_op` restores to `lhs.type`, which is the *reconciled* signless type at this point, not the authored si32/ui32. Either restore the authored type (preferred — see my other comment on the unsigned case) or reword the comment so it does not mislead future maintenance.

ptodsl/ptodsl/_scalar_adaptation.py
```diff
@@ -246,7 +260,13 @@ def _materialize_runtime_literal(value, anchor_type):
     if kind == "index":
         return arith.ConstantOp(anchor_type, int(value)).result
 
-    return _materialize_integer_literal(anchor_type, value)
+    # Runtime arithmetic is lowered through builtin ``arith`` integer ops,
+    # whose operands are signless.  Materialize literals directly as signless
+    # constants instead of constructing a signed value and immediately
+    # stripping it again (the latter used to print malformed ``siN to iN``
+    # casts for expressions such as ``num_tokens + 3``).
+    signless_type = _signless_integer_type(anchor_type)
```
Zhendong404, line 272: `int(value)` narrows the accepted literal forms: the previous path went through `_materialize_integer_literal` → `_parse_integer_value`, which also accepts integer strings (e.g. `"0x10"`). If string literals are still part of the surface, this is a small behavior regression; suggest `raw = _parse_integer_value(value, target_type=anchor_type)` here to keep parity.

ptodsl/tests/test_issue_1405_signed_runtime_arithmetic.py
```diff
@@ -0,0 +1,41 @@
...
+    assert "func.func @signed_runtime_arithmetic_probe" in text
+    assert "arith.addi" in text
+    assert "arith.floordivsi" in text
+    assert "si32" in text
+    assert "i32" in text
+    # The source of every signless arithmetic constant must be i32 itself;
+    # the malformed form was `... %c3_i32 : si32 to i32`.
+    import re
+
+    assert not re.search(r"%c\w+ = builtin\.unrealized_conversion_cast .*: si32 to i32", text)
```
Zhendong404, line 41: Test coverage suggestions: (1) add an unsigned (`pto.ui32`) case covering add/floordiv/mod/compare/min/max — that is where the reconcile rule can silently flip op selection to signed (`divui`→`floordivsi`, `ult`→`slt`); (2) this negative regex is tied to one specific malformed shape (`%c*` name, si32→i32); a more robust check would catch any `unrealized_conversion_cast` whose declared source type mismatches the defining constant, or at least generalize to `si\d+ to i\d+`; (3) `import re` belongs at module top; (4) `assert "si32" in text` / `"i32" in text` are nearly tautological and can be dropped or tightened.

ptodsl/ptodsl/_scalar_adaptation.py
liuzidi: Fixed in 0331b1e1b. Reconciliation now returns the authored integer type separately from the signless arithmetic operands. Runtime op selection uses that authored type, so ui32 with a literal lowers to divui/remui and unsigned comparison predicates, while MLIR arithmetic operands remain signless. Results are restored to the authored uiN/siN type. A dedicated ui32 regression now covers add, floor division, remainder, comparison, min, and max.

ptodsl/ptodsl/_scalar_adaptation.py
liuzidi: Fixed in 0331b1e1b. The result is now restored using the authored type captured before signless reconciliation, rather than the reconciled lhs.type. The comment is accurate with the updated implementation.

ptodsl/ptodsl/_scalar_adaptation.py
liuzidi: Fixed in 0331b1e1b. _materialize_runtime_literal now uses _parse_integer_value(value, target_type=anchor_type) before creating the signless constant, preserving integer strings such as "0x10".

ptodsl/tests/test_issue_1405_signed_runtime_arithmetic.py
liuzidi: The unsigned regression was added in 0331b1e1b and checks divui, remui, cmpi ult, maxui, and minui. The malformed-cast assertion remains focused on the exact siN-to-iN constant-cast shape that triggered this PR; authored-type propagation now covers the broader semantic issue. The top-level import re and assertion cleanup are minor follow-ups, and can be included in a subsequent cleanup if preferred.

